PR #241, opened 2023-02-06
Some tweaks to menus

keinengott:
I know that I saw that this will be overhauled eventually. In the meantime, included are some changes that make some of the menus easier to use. In particular, you can use items now by just hitting 'S' twice instead of having to cursor the item, then cursor the dragoon. Additionally, this makes the additions submenu behave more like the other menus in terms of navigation. I will do the other menus if you want, but these were the ones I found myself in most.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,10 +32,6 @@ hs_err_pid*
 
 
 
-/.idea/
-
-
-
 .gradle
 /build/
 /out/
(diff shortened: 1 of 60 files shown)


LordMonoxide (changes requested):
Usage issues:
- No keyboard support on goods menu
- "Use item" menu scrolls too far, allowing you to try to use a blank item, causing a crash
- When using an item, you have to press D to go back, instead of escape
- When using an item that heals all characters, pressing left or right causes a crash
- No key support for config menu (any key closes the menu)
- Shop item/equipment lists scroll too far
- No sound effect when selecting buy/sell/carried/leave in shop menu

PR #259, opened 2023-02-21
Rename and refactor entities relating to battle preloading entities with a focus on additions

Ink230:
## Explanation
Deduced the classes pertaining to battle preloading and initialization and provided them with related names.
Refactored battleMapSelectedAdditionHitProperties to be more readable and extensible for future edits.
Tried to stay inline with the current naming schemes present.

### BattlePreloadedEntities
BattlePreloadedEntities was the best name I could come up with for this class. We have a lot of miscellaneous collections of data present. From testing, it seems everything in this grouping is preloaded before the battle screen updates. Which is inline with the parent callers both initializing the battle and also beginning to load (see preload) assets and scripts. Thus, the name is implying that this is a collection of entities that can and must be preloaded before the battle screen updates. For this, it seems like an apt name.

Wasn't sure if it was apart of the scheme to keep the _1f8003f4 suffix to battlePreloadedEntites_1f8003f4.

## Tested on main
• All additions, complete and incomplete
• Swapping additions out between battles
• Swapping characters out
• Adjusted hit property parameters to ensure they are set, get and retrieved in both data structures through the event system

### DIFF
--- a/src/main/java/legend/game/combat/types/BattlePreloadedEntities.java
+++ b/src/main/java/legend/game/combat/types/BattlePreloadedEntities.java
@@ -9,10 +9,10 @@ import legend.game.types.MrgFile;
 import java.util.Arrays;
 
 /** 0x18cb0 bytes */
-public class BattleStruct18cb0 {
+public class BattlePreloadedEntities {
   public EncounterData38 encounterData_00;
   /** 3 slots for chars, 3 slots for dragoons */
-  public final AdditionStruct100[] _38 = new AdditionStruct100[0x100];
+  public final AdditionHits[] additionHits = new AdditionHits[0x100];
   /** This reference is only valid while it's loading */
   public MrgFile stageMrg_638;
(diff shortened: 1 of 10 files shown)


src/main/java/legend/game/combat/types/BattlePreloadedEntities.java
```diff
@@ -22,21 +22,21 @@ public class BattleStruct18cb0 {
 
   public final Rendering1298[] _9ce8 = new Rendering1298[3];
 
-  public BattleStruct18cb0() {
-    Arrays.setAll(this._38, i -> new AdditionStruct100());
+  public BattlePreloadedEntities() {
```
LordMonoxide, line 25: Add size and offset suffixes back on classes and properties (these are extremely important debugging and comparing code/memory with retail)

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #277, opened 2023-02-26
Auto Alert Icon - Quality of life improvement

LatchGameDev:
Made it so the alert icon is shown briefly when selecting a script from the Smap debugger. This makes it faster to identify the relevant script the user is looking for.


Ink230:
1) Rebase your fork branch to be at the tip or closer to the head of the current main branch. 150 commits is a lot. Then test it again to make sure it works.
image

2) I'm currently getting this error, even after having fresh files. This does not occur on main.
image

3) Other than that it works AND follows the XYZ of the script✔
image

### Questions

1. It does not work with the player submap object. Or the icon at least does not appear. Is this intentional? Are there any other submap objects it will not work with?

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -133,6 +139,33 @@ private String getSobjName(final int index) {
     return "Script %d".formatted(index);
   }
 
+  private void delayedSetToOriginalState() {
+    this.setToOriginalState();
+  }
+
+  private void setToOriginalState() {
+    LOGGER.error(this.alertIconOriginalState);
```
Ink230, line 147: LOGGER not apart of the class type w/ markers

Ink230:
image

These errors are output to console but the game does not crash.
Should be a warning or a regular log if no crash involved. 
If its suppose to crash, make it crash.

Not sure what purpose they have being logged.

LatchGameDev:
> image
> 
> These errors are output to console but the game does not crash. Should be a warning or a regular log if no crash involved. If its suppose to crash, make it crash.
> 
> Not sure what purpose they have being logged.

Sorry about that, the logged errors have accidently been committed. There were just easy to see messages to myself when I was debugging/coding which should have been removed before it left my local branch. 

Let's put the code review on pause. Still got a bit of work to do on this. Will follow up with a DM to you (Ink) for additional questions.

LatchGameDev:
> 1. It does not work with the player submap object. Or the icon at least does not appear. Is this intentional? Are there any other submap objects it will not work with?

I was testing it on maps where it did work with the player, so no it's not intentional. My code just turns it on or off so this should have been an existing issue before I made my changes. I'll see if I can track down what's going on.

LordMonoxide:
> > 1. It does not work with the player submap object. Or the icon at least does not appear. Is this intentional? Are there any other submap objects it will not work with?
> 
> I was testing it on maps where it did work with the player, so no it's not intentional. My code just turns it on or off so this should have been an existing issue before I made my changes. I'll see if I can track down what's going on.

It won't work on any map where Dart can normally have the indicator appear over his head because the scripts are constantly setting it to false

LatchGameDev:
> It won't work on any map where Dart can normally have the indicator appear over his head because the scripts are constantly setting it to false

Ah ok since that is the case then I think this is ready for a re-review / merge.

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -133,6 +138,32 @@ private String getSobjName(final int index) {
     return "Script %d".formatted(index);
   }
 
+  private void delayedSetToOriginalState() {
+    this.setToOriginalState();
+  }
+
+  private void setToOriginalState() {
```
Ink230, line 145:
```
if(this.sobj == null) return;

boolean isSelected = !this.alertIconOriginalState;
this.alertIcon.setSelected(isSelected);
this.sobj.showAlertIndicator_194 = isSelected;
```

then u can remove turnon and turnoff methods. See comment below.

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -250,6 +286,7 @@ public void collider1000Click(final ActionEvent actionEvent) {
   public void showAlertIconClick(final ActionEvent actionEvent) {
     if(this.sobj != null) {
       this.sobj.showAlertIndicator_194 = this.alertIcon.isSelected();
+      this.alertIconOriginalState = this.alertIcon.isSelected();
```
Ink230, line 289: If you can use this.alertIcon.isSelected() to find true or false on the state of the button...why do you need a private boolean property alertIconOriginalState? And elsewhere like line 195.

Ink230 (changes requested):
image

Looking good. It's able to track entities as they move. ✔

Left some comments for the code itself.

Additional stuff that probably doesn't need to block the merge but should be considered.

A toggle to:
• close all alert icons if the smap debugger is closed
• retain alerts on specific screens (after screen leaving and re-entering)

Could be useful in the future for future modders.

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -133,6 +138,32 @@ private String getSobjName(final int index) {
     return "Script %d".formatted(index);
   }
 
+  private void delayedSetToOriginalState() {
```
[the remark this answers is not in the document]
LatchGameDev: I originally added it for code readability. I was concerned me or someone else would wonder why the function call was seemingly delayed. Then once they traced back the call they would see the timeout executor. I also considered leaving a comment but I'd rather have it communicated in the code itself. That was my reasoning at the time but I'll remove the unnecessary function as requested.

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -133,6 +138,32 @@ private String getSobjName(final int index) {
...
+  }
+
+  private void turnOnAlertIcon() {
+    if(this.sobj != null) {
+      this.alertIcon.setSelected(true);
+      this.sobj.showAlertIndicator_194 = true;
+    }
+  }
+
+  private void turnOffAlertIcon() {
```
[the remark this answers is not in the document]
LatchGameDev: That was for code readability but I removed the turn off / turn off function and put them inside the setOrClearAlertIcons() function as requested. 

` private void setOrClearAlertIcons() {
    if(this.alertIconOriginalState) {
      if(this.sobj != null) {
        this.alertIcon.setSelected(true);
        this.sobj.showAlertIndicator_194 = true;
      }
    } else {
      if(this.sobj != null) {
        this.alertIcon.setSelected(false);
        this.sobj.showAlertIndicator_194 = false;
      }
    }
  }`

src/main/java/legend/game/debugger/SmapDebuggerController.java
LatchGameDev: Because I was updating the UI I needed something to keep track of the true state. But if I'm no longer keeping the UI updated as you're suggesting then yes I should be able to eliminate the private boolean property alertIconOriginalState.

src/main/java/legend/game/debugger/SmapDebuggerController.java
LatchGameDev: This to me just looks like a toggle. Like it will flip the icon from whatever state it was before. Looking at the code closer there are certain instances where I just immediately want to turn off the icon. Like when the user is quickly flipping between the objects and the icon isn't being shown for the full amount of time.

src/main/java/legend/game/debugger/SmapDebuggerController.java
```diff
@@ -164,7 +196,11 @@ private void displayStats(final int index) {
     this.collide800.setSelected((this.sobj.flags_190 & 0x800_0000) != 0);
     this.collide1000.setSelected((this.sobj.flags_190 & 0x1000_0000) != 0);
 
-    this.alertIcon.setSelected(this.sobj.showAlertIndicator_194);
+    this.alertIconOriginalState = this.sobj.showAlertIndicator_194;
+    this.turnOnAlertIcon();
+
+    final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+    executorService.schedule(this::delayedSetToOriginalState, 1000, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
LatchGameDev: See https://github.com/Legend-of-Dragoon-Modding/Legend-of-Dragoon-Java/pull/277#discussion_r1161351528

LatchGameDev:
Disregard previous comments that you may have seen. I deleted them and rolled back my local copy. I'll try to clean up the code where I can and list the reasons for each change once the rewrite is done.

LatchGameDev:
@Ink230  As I started implementing your suggested changes I kept coming across the reasons I made it this way in the first place. So I've cleaned up the code a little bit and commented the code. Basically I'm responding to the GitHub comments with code comments. This is so you can see my thought process / reasons why I did what I did. In this instance I find this easier/cleaner then making a bunch of github comments here. I can remove the code comments before it gets merged to main since comments tend to eventually get outdated.

Ink230:
Closing until updated.

PR #282, opened 2023-02-26
Small EventManager Redesign

Ink230:
## Proposal & Explanation

• Permit Events and EventManager to be less coupled to the decompiled /game/ code (and breaking changes)
• Provide cleaner Event posting within the decompiled /game/ effort (with room to be more properly hooked in)
• Provide an EventProvider 
• Permit more mod events to be created without heavy tech debt

The prior implementation of the EventManager was deeply coupled to the source code and was missing some event utility.

• Side-effects of the event were being applied within the /game/ code and not contained to the specific Event itself
• Event Providers had no utility to ensure a post-event call of the event's changes 
• Event Providers had no cleanup utility

## The benefits

• /game/ source code can have massive breaking changes without requiring changes in /modding/events/ types
• Events can guarantee their effects on their own (no return needed), if they exist, without requiring additional event handling code in non-/modding/ folders of the source code
• These effects are self-contained to the relevant Event type
• Modders don't have to worry about it for themselves, it is automatically applied
• Kept the old behaviour so nothing will break while other events are updated to match
• Being an EventProvider, is at the moment, an optional implementation

## Future benefits

• postEventAndApply events can more easily be injected through reflection instead of injecting inline in the source file, meaning events do not need to require additional coupled lines in files (or any other non-intrusive injection method)
• applyEvent behaviour could be enabled to be overwritten by modders for even greater flexibility (previously not possible)
• Event Providers and postEvent calls are more descriptive with their context signatures to aid in future /game/ refactors

## Example

• AdditionHitEvent was renamed to BattleMapActiveAdditionHitPropertiesEvent

• I added the context variables that would make this event useful and fully complete
• Since it can guarantee its own side-effects now and in the future, its placement in the code was incorrect
• Since the above, the **additional (and duplicated) code added to account for its placement was removed**

• It was moved to its proper placement where it contextually matches its usefulness
• It can be moved anywhere that matches this event's context, refactors or not, without relying on tightly coupled code following it

## EventListeners

EventListeners are completely blind to the design change and are not impacted by it (just as it should be). However, the rename is a breaking change, although distinct to itself.

This also enables us to change the behaviour of certain EventProviders in response to other breaking changes without already made mods becoming broken in the future.

## AdditionHitEvent Refactor

• Contextually relevant to the event
• More usable now, with a signature unlikely to be reduced with future changes
• Included battle in the name for contextual scope

## Notes
• The interface guarantees implementation, thus error handling is more elegant in the EventManager for EventProviders
• This could be a future refactor for the EventListeners to make exception handling redundant all together in there
• applyEvent can either call into existing /game/ methods or provide its own logic from the provided event post (and with private accessors not showing to the modder (ie: TablePtr addresses))
• The changes are dead simple but pretty powerful

• I'll refactor all the other events in the same way if this is a direction desired

*Sorry for this being so long but it does affect design :]*

TL:DR
  applyEvent
  • logging
  • external resource cleanup
    • we can give them access to a resource of the game and then ensure its gone from their touches afterwards
  • ensuring some OOP entity is released/not locked
  • ensuring some OOP entity is left in the right state
  • dealing with guaranteed side-effects the modder doesn't need to be concerned with
    • updating UI/screens to reflect the change the modder made
    • recalculating indices, memory...etc
    • calling other OOP entities to adjust them as needed


Ink230:
The PR opens the question of wider mod support. There are still mod design level questions to resolve before committing to a mod design implementation or expanding more modding events.

The goal is to get community feedback from potential modders, distill what we need for the modding API surface, and then correct game code to facilitate the needed modding design.

PR #272, opened 2023-02-25
additional debugger features

Zychronix:
auto advance text, fast text speed, "auto" charm potion, fix combat debugger not displaying full HP stat

PR #301, opened 2023-03-02
Added Auto Charm Potion effect to World Map

Ink230:
Encounters still triggered on the World Map previously. This corrects that from the original PR #272.

### DIFF
--- a/src/main/java/legend/game/WMap.java
+++ b/src/main/java/legend/game/WMap.java
@@ -5162,7 +5162,7 @@ public class WMap {
     //LAB_800e3780
     //LAB_800e3794
     final WMapAreaData08 area = areaData_800f2248.get(areaIndex_800c67aa.get());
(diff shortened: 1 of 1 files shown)


src/main/java/legend/game/WMap.java
```diff
@@ -5161,7 +5162,9 @@ public static void handleEncounters(final int encounterRateMultiplier) {
     //LAB_800e3780
     //LAB_800e3794
     final WMapAreaData08 area = areaData_800f2248.get(areaIndex_800c67aa.get());
-    encounterAccumulator_800c6ae8.add(area.encounterRate_03.get() * encounterRateMultiplier * 70);
+    if (!Config.autoCharmPotion()) {
```
LordMonoxide, line 5165:
```suggestion
    if(!Config.autoCharmPotion()) {
```

LordMonoxide (changes requested):

PR #302, opened 2023-03-02
Added Auto Run to the Debugger

Ink230:
## Explanation

Adds a toggle to the Debugger to Auto Run. It works by forcing the gamestate to include the flag for circle (or 0x40) at all times in the directional input gamestate handler. It handles the state by toggling the setting and on button or controller presses.

## Auto Run Config Dilemma 
• Config *seems* to be the place to put the autoRunHandleInput state manager
• Config gains a dependency import for _800bee90 (not always great for traditional Config classes)
• "Unavoidable" with how input is handled distinctly between controllers and keyboards outside of scripts
• Its the above with a one-liner state handler or 3 lines of duplicated code in 4 places ✔

• If the former is deemed acceptable, a state handler for each config option might be acceptable to avoid similar cases of duplicated code, annoying refactors, and/or to keep the Config code in the Config class and not polluting other types (self-contained)

### _800bee90 => inputDpadButtonsActionState_800bee90
• World movement
• Field movement
• Battle UI navigation
• Running
• Text dialogue popup UI navigation

Aka, everything largely controlled by the Dpad buttons. Exception is "running", which uses both a Dpad input and a Face button press.

### _800bee94 => inputFaceButtonsActionState_800bee94
• Field open menu
• Field interact (NPC, chest, ladders)
• Confirmation in battle UI, battle UI menus
• O return confirmation in battle UI, battle menus
• UI, UI Menus like the action bar, additions, dadditions, items, EXP / drops
• World map camera controls 
• World map location menu selection navigation & confirmation 

Aka, everything largely controlled by the Face buttons. World map location menu is an exception.

### _800bee98
Is this used anywhere? I removed all references of it in the code and warped around from Disc 1 to Disc 4, doing cutscenes, minigames and various interactions through menus and battle. Nothing broke. Did not make a commit with this change yet since I can't decipher the scripts.

### DIFF
--- a/src/main/java/legend/core/Config.java
+++ b/src/main/java/legend/core/Config.java
@@ -14,7 +14,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 
-import static legend.game.Scus94491BpeSegment_800b._800bee90;
+import static legend.game.Scus94491BpeSegment_800b.inputDpadButtonsActionState_800bee90;
 
 public final class Config {
   private Config() { }
@@ -163,7 +163,7 @@ public final class Config {
 
   public static void autoRunHandleInput() {
     if(autoRun()) {
-      _800bee90.or(0x40);
+      inputDpadButtonsActionState_800bee90.or(0x40);
     }
   }
 
--- a/src/main/java/legend/core/Config.java
+++ b/src/main/java/legend/core/Config.java
@@ -14,6 +14,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 
+import static legend.game.Scus94491BpeSegment_800b._800bee90;
+
 public final class Config {
   private Config() { }
 
@@ -40,6 +42,7 @@ public final class Config {
     properties.setProperty("fast_text_speed", "false");
     properties.setProperty("auto_advance_text", "false");
     properties.setProperty("auto_charm_potion", "false");
+    properties.setProperty("auto_run", "false");
   }
 
   public static int windowWidth() {
@@ -134,7 +137,7 @@ public final class Config {
     properties.setProperty("fast_text_speed", String.valueOf(!fastTextSpeed()));
   }
 
-  public static  boolean autoAdvanceText() {
+  public static boolean autoAdvanceText() {
     return readBool("auto_advance_text", false);
   }
 
@@ -142,12 +145,26 @@ public final class Config {
     properties.setProperty("auto_advance_text", String.valueOf(!autoAdvanceText()));
   }
 
-  public static  boolean autoCharmPotion() {
+  public static boolean autoCharmPotion() {
     return readBool("auto_charm_potion", false);
   }
 
   public static void toggleAutoCharmPotion() {
     properties.setProperty("auto_charm_potion", String.valueOf(!autoCharmPotion()));
+  }
+
+  public static boolean autoRun() {
+    return readBool("auto_run", false);
+  }
+
+  public static void toggleAutoRun() {
+    properties.setProperty("auto_run", String.valueOf(!autoRun()));
+  }
+
+  public static void autoRunHandleInput() {
+    if(autoRun()) {
+      _800bee90.or(0x40);
+    }
   }
 
   public static int getBattleRGB() {
--- a/src/main/java/legend/game/SMap.java
+++ b/src/main/java/legend/game/SMap.java
@@ -216,8 +216,8 @@ import static legend.game.Scus94491BpeSegment_800b._800bd7b8;
 import static legend.game.Scus94491BpeSegment_800b._800bda08;
 import static legend.game.Scus94491BpeSegment_800b._800bdc34;
 import static legend.game.Scus94491BpeSegment_800b._800bdd24;
-import static legend.game.Scus94491BpeSegment_800b._800bee90;
-import static legend.game.Scus94491BpeSegment_800b._800bee94;
+import static legend.game.Scus94491BpeSegment_800b.inputDpadButtonsActionState_800bee90;
+import static legend.game.Scus94491BpeSegment_800b.inputFaceButtonsActionState_800bee94;
 import static legend.game.Scus94491BpeSegment_800b._800bee98;
 import static legend.game.Scus94491BpeSegment_800b.afterFmvLoadingStage_800bf0ec;
 import static legend.game.Scus94491BpeSegment_800b.combatStage_800bb0f4;
@@ -3471,8 +3471,8 @@ public final class SMap {
 
   @Method(0x800e3d68L)
   public static void clearJoypadInput() {
-    _800bee90.set(0);
-    _800bee94.set(0);
+    inputDpadButtonsActionState_800bee90.set(0);
+    inputFaceButtonsActionState_800bee94.set(0);
     _800bee98.set(0);
   }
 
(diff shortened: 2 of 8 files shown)


Ink230:
Meeting with LordMonoxide and post-meeting investigation. These are notes to reconvene on the matter in the future so as to not have to re-learn the input systems again.

## Preamble

• The current way running works is by checking to see if Circle (0x40L) is set in a bit flag
• Movement is ultimately controlled at FUN_800de1d0 via scripts
• SMap and WMap handle running differently and also handle encounters differently
• A script is checking for this flag to be set to initiate the faster movement speed
• joypapInput_xxxx is the Java source layer for handling input
• _800bee90, _800bee94, _800bee98 are the original addresses where 3 types of input systems are set and read by scripts

## WMap Running & Encounters
**FUN_800e10a0**
• Checks if _800bee90 has 0x40 set
• Adjusts animation for running or walking
• is the function that calls handleEncounters
• WMap.handleEncounters(int encounterRateMultiplier) takes an int, 1 or 2, based on if you are walking or running, respectively

**WMAP.handleEncounters(int encounterRateMultiplier_xxxx)**
• The argument permits your encounter rate to accrue over time by a factor of 2 if running

**WMap.processInput( **
• checks for 0x40L being set in joypadInput_8007a39c
• If set, it adjusts the correct gamestate flags to provide the correct running animation and direction via sp4 (- vs +)

## SMap Running & Encounters
• Nothing in SMap controls or checks for if you are running. All calculations are independent and only act indirectly on your movement speed

**SMap.handleEncounters()**
• Note the lack of an encounterRateMultiplier compared to WMap's handleEncounter

**SMap.handleEncounters()**
• Uses encounterMultiplier_800c6abc.get() to see what your multiplier should be
• In SMap this is set to 1 or 4 for submaps (compared to WMaps 1 or 2) *based on the distance traveled* via hasPlayerMoved() in SMap


## Scus94491BpeSegment
**gameLoop()**
• joypadInput_8007a39c.setu(_800bee90.get()); sets the 0x40L flag to joypadInput
• This is later used only for correct animation and is updated on each game loop

## Scus94491BpeSegment_8007
• 3 main input properties
• Used for the joypadInput java source layer inputs
• joypadPress_8007a398
• joypadInput_8007a39c
• joypadRepeat_8007a3a0
• In the java source these are described as representing instant, press, and instant and press inputs
• This conflicts with how the internal input properties (below) are a) scripted, b) function
• Examples

joypadRepeat_8007a3a0
• Is described as what is used for enabling scrolling after a delay or similar behaviours
• If all code relating to this is commented out, scrolling behaviour still persists as normal
• The only thing that stops working is Coolon
• Therefore, this joypadRepeat seems to only be used for Coolon
• The internal script gamestate variable shares the same behaviour, _800bee98

 joypadPress_8007a398
• Is described as being for pressing a button down and holding it
• With this disabled, face buttons do not work in the field or in menus
• This is the same behaviour as the internal script variable, _800bee94

joypadInput_8007a39c
• Is described as being for pressing a button and only registering an instant response
• With this disabled, all Dpad buttons do not work in all screen states
• Similar behaviour to the internal script variable, _800bee90

## Scus94491BpeSegment_800b
• 3 internal script variables (in order of joypad)
• _800bee94: When disabled all Face buttons cease to work + running becomes impossible (even w/ controller input)
• _800bee90: When disabled all Dpad buttons cease to work
• _800bee98: When disabled Coolon menu navigation ceases to work

Scripts check the flags for these variables.

The behaviours of these three are similar to the joypad properties. The joypad properties gained their behaviour and key binds from how these internal states behave.

## 2 Interpretations

1 - Instant, Press, Repeat
The joypad properties describe the internal script inputs as these three typical input systems for handling different input behaviours.
This does not match the internal script variable behaviours.
It does match how input would have been setup or intended to be used in the 90s (and today).

2 - Distinct Hardcoded Values from Behaviour Testing
The internal script inputs are described mainly from their resulting behaviour when commented out / removed from script access.
_800bee90 is described as being for the Dpad inputs and a special case for running with 0x40.
_800bee94 is described as being for the Face inputs.
_800bee98 is described as being for and ?limited? to Coolon.

This matches the internal script behaviour of the variables.
This does not match with the existing definition for the joypad properties.

Either:
• There are internal script checks ensuring a button like 040 being set in _800bee90 is not active during certain states (menus)
or
• The 3 inputs are not used consistently (_800bee90 uses Dpad for submap but _800bee94 uses Dpad in menus)

Why is it weird?

Because Face buttons only concern themselves with the *supposedly described* button press event. While movement and navigation only concern themselves with the described button instant event. 

This is odd, as the behaviour should be mapped across both systems so that Triangle can be pressed on instant or on press, instead of just on press. Likewise, Dpad inputs would only be able to be pressed on instant and not on press. And this does not match the behaviour seen.

Further, _800bee98 does nothing for any action that resembles "repeat". And is never called by the script engine (can break point the set or get in GameVarParam, or remove them, to no crash or freeze).

Further, it is weird that with _800bee90 0x40 being set 24/7 has no input registered in menus. Meaning, the menu state intentionally is not checking _800bee90's 0x40 flag or is intentionally ignoring it (or all face button inputs) and only listens to Face button inputs on _800bee94. This is *weird*.

## Explanations

1 - The original decompilation did not capture the full input implementation (as described by Monoxide)
• This can explain _800bee98 being unused or unaccounted for 

2 - The original programmers for whatever reason (inexperience, limitations in global shared states and checks) hijacked the 3 input event flags to conform to what they needed with *hardcoded values and distinct behaviour respective to each input system*
• This can explain why both 1 & 2 are "correct" and why SMap did not receive a joypad java source layer for running

## PR concerns
The original motivation for the way the auto run was coded was based on the logically deduced behaviour of interpretation 2. This was decided as a bad move, as described below.


## Decisions
Monoxide decided that whatever current implementation that makes running work in submaps is hacky. 

The goal is to restore the original input system from decompilation. 

The hope is that the values are not used as distinct hard coded values for separate instances and do not have a 'scripting layer' that wraps the traditional instant, press, and repeat functionality all over again and applies it distinctly to each of the 3 internal state variables as the current behaviour suggests it does. 

The hope is that the traditional input, press, repeat input system is implemented in a standard way and was not decompiled fully. Specifically to this PR, the hope is that there is a well designed and thought out input system to handle the stick behaviour of auto-running, instead of it being a LGTM job squeezed into the cracks of a hijacked input system.

As for the PR, it was decided that the only way for auto-run in field maps to exist is to have the 0x40 flag set at all times on _800bee90. This was initially not a concern, as _800bee90 is only checked for Face inputs in the submap, while other menus use _800bee94 for Face button inputs. Thus, if 0x40 on _800bee90 is decided to only be used for running (as Dpad inputs are decided to be used only for movement or navigation), there seemed to be no issue. 

However, because a) the original input code is not present and b) the insecurity of having a button pressed 24/7 could lead to future issues if the submap only input of _800bee90 is ever overloaded with functionality for the circle button in submaps and not just remaining specifically for "running", the means of the PR were deemed unsuitable.

Note: I did not foresee a situation where the circle button would be given additional behaviour on submaps other than to toggle on running. I'm not sure what future behaviour would overload this circle button behaviour. Monoxide can explain in greater detail if needed in the future.

## Future issues
• If the input, press, repeat system is believed to be true by the decompilation, then a major refactor of how inputs are handled and passed to the internal script variables must be done so that the button mappings translate better to that paradigm. As the current button mappings and internal script mappings do not make sense in this context.

• If the alternative interpretation is true, then the way scripts handle inputs is a larger mystery and should probably be refactored anyways for future expansion.

The end result is the same, restore the original PSX input handling.

Issue #294, opened 2023-02-28
Item and equip lists show cursor simultaneously


theflyingzamboni:
Both columns in the item/equipment list have the cursor showing regardless of which column the player is navigating, making it impossible to tell which list is being navigated. The cursor should only appear in whichever column is being actively navigated.

![image](https://user-images.githubusercontent.com/55669985/221749639-000ad375-81c9-4ae3-9df4-a8bd2a494072.png)

Ink230:
This is also present in the PR #268

PR #268, opened 2023-02-25
New Input System 

LatchGameDev:
- Support for over 1800 different controllers
- F9 Will let you change what is the active controller
- Automatically connects to the controller if only one is present 
- Removed old input to hex code 
- Replaced some input hex checks in the game with the new system
- Modified screens to use the new input

Example Checks:
`if(Input.pressedThisFrame (InputAction.BUTTON_NORTH)) { }
 if(Input.pressedWithRepeatPulse(InputAction.DPAD_DOWN)) { }
 if(Input.releasedThisFrame(InputAction.BUTTON_SHOULDER_LEFT_2)) { }
`

Also created event versions of these for the screens.

PR #303, opened 2023-03-02
List Menu Dual Highlight Bug Fix + Shop Screen Sell DRG Alignment Fix

Ink230:
## List Menu Bug Fix - #294 

• Refactored render code to be simpler with sharing one highlighter, one selectSlot
• Refactored vertical input handling to be as straightforward as a shared state dual pane distinct menu system can be
• Added logic that was missing to the Left and Right toggles of state 

• This may have conflicts with #268 (which did not address this bug)

## Equipment => Armed

• Equipment broke out of the boundaries on the Sell => Equipment/Items menu
• Armed is the original name and Armed is used elsewhere in the Menu system
• I did not restore the original menu in the mid left of the screen but kept the one already present
• See following screenshots for details

### Original Bug
image

### What Retail Looks Like
image

### What The Fix Looks Like
image

### DIFF
--- a/src/main/java/legend/game/inventory/screens/ItemListScreen.java
+++ b/src/main/java/legend/game/inventory/screens/ItemListScreen.java
@@ -46,13 +46,11 @@ public class ItemListScreen extends MenuScreen {
   private double scrollAccumulator;
   private final Runnable unload;
 
-  private int selectedSlotEquipment;
-  private int selectedSlotItem;
+  private int selectedSlot;
   private int slotScrollEquipment;
   private int slotScrollItem;
   private int equippedItemsCount;
-  private Renderable58 equipmentHighlight;
-  private Renderable58 itemHighlight;
+  private Renderable58 highlight;
   private Renderable58 _800bdb9c;
   private Renderable58 _800bdba0;
   private int mouseX;
@@ -80,15 +78,13 @@ public class ItemListScreen extends MenuScreen {
         deallocateRenderables(0xff);
         renderGlyphs(goodsGlyphs_801141c4, 0, 0);
         recalcInventory();
-        this.selectedSlotEquipment = 0;
-        this.selectedSlotItem = 0;
+        this.selectedSlot = 0;
         this.slotScrollEquipment = 0;
         this.slotScrollItem = 0;
-        this.currentItemId = 0xff;
-        this.equipmentHighlight = allocateUiElement(0x76, 0x76, FUN_800fc824(0), FUN_800fc814(this.selectedSlotEquipment) + 32);
-        FUN_80104b60(this.equipmentHighlight);
-        this.itemHighlight = allocateUiElement(0x76, 0x76, FUN_800fc824(1), FUN_800fc814(this.selectedSlotItem) + 32);
-        FUN_80104b60(this.itemHighlight);
+        this.setCurrent(gameState_800babc8.equipment_1e8, this.equipment, 0);
+
+        this.highlight = allocateUiElement(0x76, 0x76, FUN_800fc824(0), FUN_800fc814(this.selectedSlot) + 32);
+        FUN_80104b60(this.highlight);
         this.equippedItemsCount = FUN_80104738(this.equipment, this.items, 0x1L);
         this.renderItemList(this.slotScrollEquipment, this.slotScrollItem, 0xff, 0xff);
         this.loadingStage++;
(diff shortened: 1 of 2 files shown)


src/main/java/legend/game/inventory/screens/ItemListScreen.java
```diff
@@ -124,18 +120,20 @@ protected void render() {
...
           if(MathHelper.inBox(this.mouseX, this.mouseY, 186, 40, 174, 122)) {
             if(this.slotScrollItem < gameState_800babc8.itemCount_1e6.get() - 7) {
               playSound(1);
               this.slotScrollItem++;
-              this.setCurrent(gameState_800babc8.items_2e9, this.items, this.slotScrollItem + this.selectedSlotItem);
+              this.setCurrent(gameState_800babc8.items_2e9, this.items, this.slotScrollItem + this.selectedSlot);
             }
           }
         }
+
```
LordMonoxide, line 135: Extra new lines (try to keep changes like this to a minimum to simplify file history)

src/main/java/legend/game/inventory/screens/ItemListScreen.java
```diff
@@ -152,6 +150,30 @@ protected void render() {
     }
   }
 
+  private void handleVerticalInput(final boolean isScrollingUp)
+  {
```
LordMonoxide, line 154: CS - brace on same line

src/main/java/legend/game/inventory/screens/ItemListScreen.java
```diff
@@ -198,20 +220,20 @@ protected void mouseMove(final int x, final int y) {
...
     if(this.loadingStage == 1) {
       for(int i = 0; i < Math.min(7, gameState_800babc8.equipmentCount_1e4.get() - this.slotScrollEquipment); i++) {
-        if(this.selectedSlotEquipment != i && MathHelper.inBox(x, y, 8, 31 + FUN_800fc814(i), 174, 17)) {
+        if(this.selectedSlot != i && MathHelper.inBox(x, y, 8, 31 + FUN_800fc814(i), 174, 17)) {
           playSound(1);
-          this.selectedSlotEquipment = i;
-          this.equipmentHighlight.y_44 = FUN_800fc814(i) + 32;
-          this.setCurrent(gameState_800babc8.equipment_1e8, this.equipment, this.slotScrollEquipment + this.selectedSlotEquipment);
+          this.selectedSlot = i;
+//          this.equipmentHighlight.y_44 = FUN_800fc814(i) + 32;
```
LordMonoxide, line 226: Commented code? Are these needed or should they be removed?

src/main/java/legend/game/inventory/screens/ItemListScreen.java
```diff
@@ -46,13 +46,11 @@ public class ItemListScreen extends MenuScreen {
   private double scrollAccumulator;
   private final Runnable unload;
 
-  private int selectedSlotEquipment;
-  private int selectedSlotItem;
+  private int selectedSlot;
   private int slotScrollEquipment;
```
LordMonoxide, line 50: Should the scroll variables be combined as well, or is there a reason to keep them separate?

LordMonoxide (changes requested):

LordMonoxide:
Can you repoint this to RC? I want to get it in for the recommended build

src/main/java/legend/game/inventory/screens/ItemListScreen.java
Ink230: I believe they're necessary. Each menu tracks its own independent state of how far down either is scrolled. You need something to say Equipment has scrolled down 9 items while Items has scrolled down 30 items (aka, holding state) for the way the lists are rendered / pulled in from gameState. They are basically just indices of two distinct sets. And the external side-effects of some decompiled functions are basically dependencies we're working to satisfy that use these indices.

This implementation, which was the original code, makes it easier (or possible) to navigate between the two menus freely with Left and Right inputs. Refactoring this into a better design pattern seems unjustified for the scope of the change, but may be justified if the other menus are all refactored into a better pattern.

src/main/java/legend/game/inventory/screens/ItemListScreen.java
Ink230: Unimplemented mouse support changes. Working on the implementations for a commit.

src/main/java/legend/game/inventory/screens/ItemListScreen.java
Ink230: Notice in the Goods menu that the scroll is global between both menus when using one slotScroll variable. Scrolling in one, scrolls in the other. This is not ideal behaviour.

Ink230:
Mouse input and functionality is restored. Keyboard navigation was what was worked on previously.

Added full functionality in navigation and selection for mouse and keyboard distinctly. Circle or back functionality on Discard menu, and in general, the menus currently, don't exist. This is a change outside of the scope of this PR. Selecting "No" is the only option to go back currently; there are no UI soft locks.

Controller support can be added into the keyPress method (and potentially renamed) as all actions are 1:1 to controllers. Easiest way is to add additional cases on the keyPress event. Other implementations might struggle with lack of controller support in the overall Menu system, see PR #268.

LordMonoxide (approved):

PR #319, opened 2023-03-11
Input System Tweaks

LatchGameDev:
- Corrected wrong formatting (used a %d instead of a %s) 
- If any issues with the controllers are detected it now properly goes to Keyboard-Mouse only mode
- In keyboard-mouse only mode it now doesn't attempt to look up mappings in the controller database
- Added a try block on disconnect until the source of the error can be found. Still can't reproduce it locally however normal DS4 controllers do seem to work fine.
- Minor Refactor
- Updated Readme with new controls / controller procedure

### DIFF
--- a/src/main/java/legend/game/input/Input.java
+++ b/src/main/java/legend/game/input/Input.java
@@ -236,11 +236,10 @@ public final class Input {
   private static void onControllerDisconnected(final int id) {
     try {
       if(playerOne.getControllerData().getGlfwControllerId() == id) {
-        // Player one has been disconnected
         LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected please reconnect or switch to a different controller using F9");
       }
     } catch(final NullPointerException exception) {
-      LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected");
+      LOGGER.error(INPUT_MARKER, "NULL POINTER EXCEPTION: ",exception.getMessage());
       InputControllerAssigner.reassignSequence();
     }
   }
--- a/src/main/java/legend/game/input/Input.java
+++ b/src/main/java/legend/game/input/Input.java
@@ -37,8 +37,6 @@ public final class Input {
   private static InputMapping playerOne = new InputMapping();
   private static final List<InputMapping> controllers = new ArrayList<>();
 
-  public static final String keyboardMouseOnly = "Keyboard-Mouse Only";
-
   private Input() {
   }
 
--- a/src/main/java/legend/game/input/Input.java
+++ b/src/main/java/legend/game/input/Input.java
@@ -234,13 +234,14 @@ public final class Input {
   }
 
   private static void onControllerDisconnected(final int id) {
-
-    if(playerOne.getControllerData().getGlfwControllerId() == id) {
-      // Player one has been disconnected
-      LOGGER.info(INPUT_MARKER,"Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+    try {
+      if(playerOne.getControllerData().getGlfwControllerId() == id) {
+        // Player one has been disconnected
+        LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+      }
+    } catch(final NullPointerException exception) {
+      LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected");
+      refreshControllers();
     }
-
-
   }
-
 }
--- a/src/main/java/legend/game/input/Input.java
+++ b/src/main/java/legend/game/input/Input.java
@@ -236,7 +236,7 @@ public final class Input {
   private static void onControllerDisconnected(final int id) {
     try {
       if(playerOne.getControllerData().getGlfwControllerId() == id) {
-        LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+        LOGGER.info(INPUT_MARKER, "Player 1's controller has been disconnected. Please reconnect the controller, or switch to a different controller using F9.");
       }
     } catch(final NullPointerException exception) {
       LOGGER.error(INPUT_MARKER, "NULL POINTER EXCEPTION: %s ",exception.getMessage());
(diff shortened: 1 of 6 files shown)


src/main/java/legend/game/input/Input.java
```diff
@@ -234,13 +234,13 @@ private static void onControllerConnected(final int id) {
...
-
-    if(playerOne.getControllerData().getGlfwControllerId() == id) {
-      // Player one has been disconnected
-      LOGGER.info(INPUT_MARKER,"Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+    try {
+      if(playerOne.getControllerData().getGlfwControllerId() == id) {
+        LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+      }
+    } catch(final NullPointerException exception) {
+      LOGGER.error(INPUT_MARKER, "NULL POINTER EXCEPTION: ",exception.getMessage());
```
Ink230, line 242: ```LOGGER.error(INPUT_MARKER, "NULL POINTER EXCEPTION: " + exception.getMessage());```

src/main/java/legend/game/input/InputControllerAssigner.java
```diff
@@ -86,7 +80,7 @@ public static void reassignSequence() {
 
     for(int i = 0; i < GLFW_JOYSTICK_LAST; i++) {
       if(glfwJoystickPresent(i)) {
-        LOGGER.info(INPUT_MARKER,"%d: %s (%d)",(i+1),glfwGetJoystickName(i),glfwGetJoystickGUID(i));
+        LOGGER.info(INPUT_MARKER,"Controller Id: %d - GUID: %s",i ,glfwGetJoystickGUID(i));
```
Ink230, line 83: CS:
• Space between arguments in LOGGER.info (missing 3)

Ink230:
"Keyboard-Mouse Only" should be an enum or use derived enum advantages (type safety, comparisons)

This might be a larger refactor for InputControllerData and other dependable methods

• Using hardcoded strings in more than one place indicates an enum should (probably) be used
• Variables with the same string values BUT which mean different things should be distinct enums
• ex: InputControllerData takes glfwJoystickName and glfwJoystickGUID but are given the same hardcoded string
• This is not ideal currently and could become a pain in the future if future enum-like hardcoded strings are needed for InputControllerData or other status tracking (which enums were made for)

Ink230:
1. When you make a public static final field on a type you are basically letting anyone throughout the whole code modify or change the variable, despite it only being needed in scope to the Input types and Input handling.

2. Further, if the field's intent is to never change, it is effectively a global constant variable. 

3. If the field is also required to be compared (.equals) to something, that *implies* it can have multiple values and is a stateful tracking field. 

You could make a getter and setter or a (isKeyBoardInput) method on Input to handle issues 1 and 3 but that contradicts the behaviour that issue 2 has. 

A.
However, all 3 are resolved by using an enum (no field on Input needed, Input does not need a tracking state field, no getter/setter needed). Then, you can compare the enum directly of itself with the .equals with targetControllerGUID.

This is *better* for the short circuit you have going on in the getBindings method and can clean up the "hacky" assignments in InputControllerData. With an enum, you can also make the short circuit cleaner and plan for future short circuit cases without manually having to assign all the potential values.

Whether you use a casted int enum to string or a string enum is up to your implementation.

B.
The issue of the same variable being used for two different behaviours is still an issue and is not good practice.

C.
Without a lot of refactoring in the controller db mapping functionality the short circuit approach is the best. However, consider the placement of the short circuit in getBinding(). Should it be there or in Input.refreshControllers() or even in the isEmpty controller check or somewhere else. This might permit more concise code with the option for more granular short circuiting features while reducing the short circuits complexity.

LatchGameDev:
I agree that having a hard coded string in multiple places is bad. If someone changes only one of them it changes the logic. But once I threw that in a constant/final then that problem is solved. 

I stepped up one function which gave me access to the glfwControllerId which is -1 when no controllers are connected. 

I don't want to use an enum for a simple short circuit.

Lastly there are things that I want to refactor about the Input System. There is some redundant code I haven't eliminated yet but I was trying to not make large changes since we are so close to the recommended build. I didn't want to introduce a last minute bug.

src/main/java/legend/game/input/Input.java
```diff
@@ -234,13 +234,13 @@ private static void onControllerConnected(final int id) {
   }
 
   private static void onControllerDisconnected(final int id) {
-
-    if(playerOne.getControllerData().getGlfwControllerId() == id) {
-      // Player one has been disconnected
-      LOGGER.info(INPUT_MARKER,"Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+    try {
+      if(playerOne.getControllerData().getGlfwControllerId() == id) {
+        LOGGER.info(INPUT_MARKER, "Player 1 has been disconnected please reconnect or switch to a different controller using F9");
```
LordMonoxide, line 239:
```suggestion
        LOGGER.info(INPUT_MARKER, "Player 1's controller has been disconnected. Please reconnect the controller, or switch to a different controller using F9.");
```

src/main/java/legend/game/input/Input.java
```diff
@@ -234,13 +234,13 @@ private static void onControllerConnected(final int id) {
...
-
-    if(playerOne.getControllerData().getGlfwControllerId() == id) {
-      // Player one has been disconnected
-      LOGGER.info(INPUT_MARKER,"Player 1 has been disconnected please reconnect or switch to a different controller using F9");
+    try {
+      if(playerOne.getControllerData().getGlfwControllerId() == id) {
+        LOGGER.info(INPUT_MARKER, "Player 1's controller has been disconnected. Please reconnect the controller, or switch to a different controller using F9.");
+      }
+    } catch(final NullPointerException exception) {
+      LOGGER.error(INPUT_MARKER, "NULL POINTER EXCEPTION: %s ",exception.getMessage());
```
LordMonoxide, line 242:
```suggestion
      LOGGER.error(INPUT_MARKER, "NPE on controller disconnection", exception);
```

LordMonoxide (approved):

PR #321, opened 2023-03-11
Implement "Game PAUSED" and Additional Joystick code

dragoonsouls:
Now when game pauses, the title bar will show "PAUSED", also there is some additional code provide by Latch for working around with the DS4Windows problem.
Files changed:
GPU.java
Scus94491BpeSegment.java
Input.Java
Also is adding a change to the controls.txt that actually need to be updated. (This is the old one, before the big changes in the Input.java).

### DIFF
--- a/src/main/java/legend/game/inventory/screens/TooManyItemsScreen.java
+++ b/src/main/java/legend/game/inventory/screens/TooManyItemsScreen.java
@@ -135,6 +135,10 @@ public class TooManyItemsScreen extends MenuScreen {
               playSound(1);
               this.invScroll--;
               this.renderable_8011e204.y_44 = this.FUN_8010f178(this.invIndex);
+            } else if(this.invScroll == 0) {
+              this.invScroll = slotCount - 7;
+              this.invIndex = 6;
+              this.renderable_8011e204.y_44 = this.FUN_8010f178(this.invIndex);
             }
           }
         }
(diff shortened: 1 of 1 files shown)


src/main/java/legend/core/gpu/Gpu.java
```diff
@@ -427,6 +428,9 @@ public void run() {
 
       final float fps = 1.0f / ((System.nanoTime() - this.lastFrame) / (1_000_000_000 / 30.0f)) * 30.0f;
       this.window.setTitle("Legend of Dragoon - FPS: %.2f/%d".formatted(fps, this.window.getFpsLimit()));
+      if(enginepaused){
```
Ink230, line 431: CS:
```){``` 
should be
```) {```

File: src/main/java/legend/game/Scus94491BpeSegment.java [hunk 2]
```diff
@@ -417,6 +417,7 @@ private Scus94491BpeSegment() { }
   public static final Int2IntMap keyRepeat = new Int2IntOpenHashMap();
 
   private static boolean paused;
+  public static boolean enginepaused;
```
Ink230, line 420: camelCase should be used for naming
```enginepaused``` should be
```enginePaused```

Ink230:
Some questions to ask as a learning exercise relating to the "Game PAUSED" feature

• Why was a new field created when paused encapsulates the same behaviour?
• Why was the new field made public instead of using a getter / setter? Is this needed?
• Does GPU.java need to know about ```Scus94491BpeSegment``` via an import? Does it already know of ```Scus94491BpeSegment``` or does it have another means of accessing ```Scus94491BpeSegment```? 
• Which way is the information flowing currently between ```Scus94491BpeSegment``` and GPU.java? Does your change adhere to that?
• hint: gameLoop(), this.window.setTitle, this.ctx.onDraw

Ink230:
For the "Additional Joystick Code"
#319 fulfills that scope

Ink230:
For the 32 commit merge history

• Look into rebasing your history to make it concise and related to your new commits
• Isolate the two commits that are meaningful so only those are displayed in the PR history
• Make sure commits have unique describing messages attached to them (your two share the same message)
• Is the first commit necessary / within the scope of your "Game PAUSED" feature? If not maybe it should be its own PR, if not covered by another PR.

PR #322, opened 2023-03-12
VRAM split with rendering decoupled from TIMs and TMDs

LordMonoxide:


Ink230:
Checked it out, didn't crash. ✔

PR #344, opened 2023-03-22
Fixed some non-bug causing code issues

theflyingzamboni:
Just a couple things I found looking for the possible weapon trail issues, plus some light renaming. Shouldn't actually change how anything works, so I'm PRing main instead of rc.

### DIFF
--- a/src/main/java/legend/game/combat/Bttl_800c.java
+++ b/src/main/java/legend/game/combat/Bttl_800c.java
@@ -3599,8 +3599,8 @@ public final class Bttl_800c {
 
   @Method(0x800cdcecL)
   public static void FUN_800cdcec(final Model124 model, final int dobjIndex, final VECTOR largestVertRef, final VECTOR smallestVertRef, final EffectManagerData6c manager, final IntRef largestIndexRef, final IntRef smallestIndexRef) {
-    short largest = 0x7fff;
-    short smallest = -1;
+    short largest = -1;
+    short smallest = 0x7fff;
     int largestIndex = 0;
     int smallestIndex = 0;
     final TmdObjTable1c v0 = model.dobj2ArrPtr_00[dobjIndex].tmd_08;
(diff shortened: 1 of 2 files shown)


src/main/java/legend/game/combat/Bttl_800c.java
```diff
@@ -3671,30 +3671,30 @@ public static WeaponTrailEffectSegment2c FUN_800cde1c(final WeaponTrailEffect3c
...
       final IntRef x2 = new IntRef();
       final IntRef y2 = new IntRef();
-      final int z = transformWorldspaceToScreenspace(s0._04.get(1), x2, y2) >> 2;
+      final int z = transformWorldspaceToScreenspace(effectSegment._04.get(1), x2, y2) >> 2;
 
       //LAB_800cdf94
-      s0 = s0._24.derefNullable();
-      for(int i = 0; i < s2._0e && s0 != null; i++) {
+      effectSegment = effectSegment._24.derefNullable();
+      for(int i = 0; i < effect._0e && effectSegment != null; i++) {
```
Ink230, line 3691: 1.  ```i < effect._0e && effectSegment != null```
Fairly non-standard way for a for loop to be conditioned. Can this be simplified (.length, count, isNull), (other variable to track changes of, not so many side-effect checks).

2. Does the loop need to exist in its current form?
• i is not used anywhere
• Can you confirm in Ghidra this is the way the loop should be constructed, it seems odd for a single linked list traversal to be made this way?
• Elsewise, a while loop is more justified for a single linked list traversal with a single null check on the node than having a full for loop.

3. Is this worth doing?
```¯\_(ツ)_/¯```
=> could be optimization potential in here w/ a cache on same coord results being reused
=> IntRefs (x1, y1, x3, y3) don't need to be included in each iteration...or do they?

theflyingzamboni: 1. ¯\\\_(ツ)_/¯ This is how it was already written. Also, `effect._0e` *is* the count of the number of effect segments.
2. Why is it odd to iterate through a list?
3. Those IntRefs do need to be included, because they're different for each effect segment.

LordMonoxide (approved):

PR #368, opened 2023-04-01
Fixes switching camera views in combat

theflyingzamboni:
Typo was writing array values to wrong address.


Ink230 (approved):
_800c6c30 exists as a property in Bttl_800c and as an import in Bttl_800f. It's all copasetic.

LordMonoxide (approved):

PR #378, opened 2023-04-02
Add encounter rate config (closes #370)

### LINKED ISSUE #370: Add config for encounter rates

LordMonoxide:


Ink230 (commented):
This is all *ok* providing BaseMod is a temporary solution and it is not going to become more overloaded.

- It is weird that BaseMod.Anything may be called from game code instead of ensuring BaseMod only affects game stuff through EventManager.INSTANCE (and ensuring any future factors only have to refactor things one way) or through registries affecting game state objects on initial load (not so for WMap or SMap which are called endlessly at runtime).

- I think if you have to add public static properties to BaseMod to affect game logic or game data a) something is wrong or b) use a fully implemented regular event to reference the state or c) add it to one of the gameState objects.

- I appreciate that BaseMod can be the place to rig together player selected options that are different to retail but are native to Severed Chains. I also get that this is totally fine for things that take effect on initial load. But having BaseMod be called directly by game code for its state through public static properties *during runtime* has to sound *not good* for a) now and b) in the future and c) future refactors (in BaseMod and in game code).

- Where do the config values live then? In the @ EventListener registerConfig(final ConfigRegistryEvent event) *and/or* in a @ EventListener defaultConfig(final DefaultConfigEvent event). With state outlined inside of either of those methods and conforms to whatever is provided in the backing type (potentially like serialization utilities...). .setEditControl will work in either case (though that might also not be a good way of doing things from within BaseMod...).

- We can add BaseMod.Property anywhere we want in the code. Modders can not add BaseMod.Property to anywhere they want in the code. Having BaseMod be "special" in its implementation seems like a bad design when we can have the same behaviour across BaseMod and regular mods. This seems like a better way to simplify things.

PS: Didn't catch on to BaseMod.INVENTORY_SIZE_CONFIG in a previous PR.

PR #381, opened 2023-04-02
Add version to unpack (closes #380)

### LINKED ISSUE #380: Add versioning to unpacker
Now that we have official builds, we can't just tell users to delete their files directory when updating.

LordMonoxide:


Ink230 (commented):
LGTM first implementation. Anything more complicated with partial version increment logic can come when we run into it unless it is very clear that some portions of files can be micro optimized easily for reduced unpacker times / deletion times.

PR #411, opened 2023-04-10
Rename, Refactor AdditionHitEvent => BattleMapActiveAdditionHitPropertiesEvent

Ink230:
### Renamed AdditionHitEvent to BattleMapActiveAdditionHitPropertiesEvent

Since,
a) it has to do with battle
b) it is mapping addition hit properties
c) it is only mapping for "Active Stats" version of hit properties.

### Refactored the event to be placed in a more useful location, with more passed event properties

Previously, you would have needed to call game state objects to get these new passed properties or done some shenanigans with the addition hit properties to identify characters, char slots, dragoon or not...etc.

Also, previously the event wasn't respecting pass by reference fundamentals, and now it does.

• additionIndex: Which addition is triggering the event
• charIndex: Which character is triggering the event
• charSlot: Which battle slot is triggering the event
• dragoon: true for if the addition is a dragoon addition, false for regular addition


Ink230:
#465 will conflict with this PR.
This will need one line, one name change in

```BattleMapActiveAdditionHitPropertiesEvent.java```

```...final BattlePreloadedEntities_18cb0.AdditionHits_100 additionHits,...```

to

```...final BattlePreloadedEntities_18cb0.AdditionHits100 additionHits,...```

LordMonoxide (approved):

PR #418, opened 2023-04-14
Fixed sustain, for some reason requires slightly more voices.

Illeprih:


src/main/java/legend/core/audio/AudioThread.java
```diff
@@ -0,0 +1,126 @@
...
+
+    for(final Voice voice : this.voicePool) {
+      voice.destroy();
+    }
+    alcDestroyContext(this.audioContext);
+    alcCloseDevice(this.audioDevice);
+  }
+
+  public void stop() {
+    this.running = false;
```
LordMonoxide, line 103: What happens if it's paused when `stop` is called? I think this will have to notify() to wake it up from the wait()

src/main/java/legend/core/audio/AudioThread.java
```diff
@@ -0,0 +1,126 @@
...
+  private boolean running;
+  private boolean paused;
+
+  /**
+   *
+   * @param frequency Amount of updates per second. Has to be a divisor of 1_000_000_000 and 44_100 (1, 2, 4, 5, 10, 20, 25, 50, 100)
+   * @param voiceCount Amount of voices that can play at ones. Retail uses 24
+   */
+  public AudioThread(final int frequency, final boolean stereo, final int voiceCount) {
+    assert 1_000_000_000 % frequency == 0 : "1_000_000_00 is not divisible by " + frequency;
```
LordMonoxide, line 37: Make both of these IllegalArgumentExceptions (think that's it) and throw something else for the other assert below

src/main/java/legend/core/audio/Offsets.java
```diff
@@ -0,0 +1,26 @@
+package legend.core.audio;
+
+final class Offsets {
+  static double[] semitone = new double[12];
+  static double[] cent = new double[100];
+  static double[] pan = new double[128];
+
+
+  static void genOffsets() {
```
LordMonoxide, line 9: You could just make this a `static {` block so the first time this class is accessed it generates them

src/main/java/legend/core/audio/Voice.java
```diff
@@ -0,0 +1,245 @@
...
+    if(adsrValue < 16) {
+      this.empty = true;
+    }
+
+    final double volume = this.channel.getVolume() * this.layer.getVolume() * this.velocity;
+
+    final short processedSample = (short)(((int)(sample * adsrValue * volume)) >> 15);
+
+    if(this.sound.isStereo()) {
+      //TODO panning
```
LordMonoxide, line 56: Is panning done? If so remove the TODO

LordMonoxide (changes requested):

src/main/java/legend/core/audio/AudioThread.java
Illeprih: Done

src/main/java/legend/core/audio/Offsets.java
Illeprih: Could cause a stutter

LordMonoxide (approved):

PR #435, opened 2023-04-22
Tiny hands

theflyingzamboni:
See LAB_80109b90 in ghidra.

In applyVertexDifferenceAnimation, previous was being set equal to current instead of previous + current.


LordMonoxide (approved):
Albert's arms are all grown up

PR #436, opened 2023-04-23
Fixed lighting on LSC translucent textured prims

theflyingzamboni:
Needed to explicitly exclude lit primitives from the uniform lighting logic branch.

Don't know how many things this affected, but Lavitz's damaged chest plate during oof is an example, now correct.

### DIFF
--- a/src/main/java/legend/game/tmd/Renderer.java
+++ b/src/main/java/legend/game/tmd/Renderer.java
@@ -173,7 +173,7 @@ public final class Renderer {
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


Ink230 (approved):
Nice PR size, keep it up. ✔
Code fine. ✔
Wonder what this breaks for what it fixes. ✔

theflyingzamboni:
Hypothetically, it breaks nothing. The uniform lighting code should always have been restricted to non-lit primitives, because it applies to rendering CTMD NLSC translucent textured primitives specifically.

PR #445, opened 2023-04-29
Fix applyBuffOrDebuff with negative values (closes #424)

### LINKED ISSUE #424: applyBuffOrDebuff with negative value
Need to confirm this, but it looks like if debuff spells are used, the negative value will blow away the turn count and set it to 255

LordMonoxide:


theflyingzamboni (approved):
Adds code what's in the assembly.

PR #448, opened 2023-04-29
Fix floating text centering (closes #407)

### LINKED ISSUE #407: Floating number centering centers as if there are leading 0s
It's easy to see when enemies do damage to you. The numbers are too far over.

LordMonoxide:


Ink230 (approved):
LGTM!

PR #450, opened 2023-04-29
Fix world map text centering (closes #215)

### LINKED ISSUE #215: World Map: Location "Service Listings" unaligned
For this screen, I believe left-alignment is best for the service listing at the top.
image

LordMonoxide:


src/main/java/legend/game/Scus94491BpeSegment_8002.java
```diff
@@ -4656,8 +4656,17 @@ public static int textWidth(final LodString text) {
...
   public static int textWidth(final String text) {
     int width = 0;
+    int currentWidth = 0;
     for(int index = 0; index < text.length(); index++) {
-      width += charWidth(text.charAt(index));
+      if(text.charAt(index) == '\n') {
+        currentWidth = 0;
+      }
+
+      currentWidth += charWidth(text.charAt(index));
```
Ink230, line 4665: is the currentWidth += needed if text.CharAt(index) points to '\n', as this width would be 0? 
Then also the currentWidth = 0 is redundant.

```
if(text.charAt(index) == '\n') {
    continue;
}
```

Saves 0.1 of a frame.

also

```
if(charWidth(text.charAt(index)) == 0) {
    continue;
}
```

might be a more general check if appropriate for the context of notdoing work with all 0 width symbols and not just '\n'.

Ink230: Also, most of the codebase uses i not index. 🧐

Ink230 (approved):
LGTM!

PR #451, opened 2023-04-29
Fixes for characters that are supposed to be oofed (closes #425)

### LINKED ISSUE #425: Modded party oddity on char select screen
With Shana modded into the party, if you remove her, "Don't Select" appears over her in the secondary char slot

image

LordMonoxide:


Ink230 (approved):
Nice fix. ✔

PR #456, opened 2023-05-01
Fixed targeting squares not being correct

### LINKED ISSUE #414: "Too late" addition square is gray instead of blue
The blue square when you press X too late never seems to appear. Even if you don't press X at all, you get the gray "too early" square.

theflyingzamboni:
When initializing the structs at the start of the addition, the pointer was being decremented one time too many at the end of the loop. Moved the decrement to the start of the loop and removed the standalone decrement before the loop.

There was another bug at L3732-3734, where the subtraction is supposed to be performed prior to division. This definitely changes some values, but I can't tell what the effect is in-game.

Closes #414


Ink230 (approved):
Discussed in Discord the precision question and ensured the line 3700 change in SEffe.java was real life.

LordMonoxide (approved):

PR #465, opened 2023-05-06
Addition overlay refactor and config/debug controls

theflyingzamboni:
All memory related to the addition overlay has been moved into ref structs, and related code refactored and named.

Also created a safer and less hacky implementation of auto addition, added auto-addition and addition overlay display controls to game options, and added options to control overlay colors to the debugger.


src/main/java/legend/game/combat/SBtld.java
```diff
@@ -118,8 +118,8 @@ public static void FUN_80109170(final FileData file) {
   public static void battlePrepareSelectedAdditionHitProperties_80109250() {
     //LAB_801092a0
     for(int charSlot = 0; charSlot < 3; charSlot++) {
-      final BattlePreloadedEntities_18cb0.AdditionHits_100 activeAdditionHits = battlePreloadedEntities_1f8003f4.additionHits_38[charSlot];
-      final BattlePreloadedEntities_18cb0.AdditionHits_100 activeDragoonAdditionHits = battlePreloadedEntities_1f8003f4.additionHits_38[charSlot + 3];
+      final BattlePreloadedEntities_18cb0.AdditionHits100 activeAdditionHits = battlePreloadedEntities_1f8003f4.additionHits_38[charSlot];
```
Ink230, line 121: Are we suppose to be removing the _

src/main/java/legend/game/combat/environment/BattlePreloadedEntities_18cb0.java
```diff
@@ -8,12 +8,18 @@
...
-/** 0x18cb0 bytes */
+/**
+ * 0x18cb0 bytes
+ */
 public class BattlePreloadedEntities_18cb0 {
   public EncounterData38 encounterData_00;
-  /** 3 slots for chars, 3 slots for dragoons */
-  public final AdditionHits_100[] additionHits_38 = new AdditionHits_100[0x100];
-  /** This reference is only valid while it's loading */
+  /**
```
Ink230, line 16: A lot of line changes just for a diff comment syntax

Ink230 (dismissed):

src/main/java/legend/game/combat/SBtld.java
theflyingzamboni: @LordMonoxide should chime in, but I think the _ might have been an early naming convention for structs, because most of our structs don't have them.

src/main/java/legend/game/combat/environment/BattlePreloadedEntities_18cb0.java
theflyingzamboni: Stupid auto-formatting. I think I changed my settings to preserve one-line docstrings, but it's not in the project file on the repo. Guess this got altered before I changed that.

src/main/java/legend/game/combat/SBtld.java
LordMonoxide: Either is fine, I usually only use a `_` if there are back to back numbers that need to be separated

src/main/java/legend/game/combat/Bttl_800d.java
```diff
@@ -883,7 +884,19 @@ public static FlowControl allocateGuardEffect(final RunningScript<? extends Batt
     effect._00.set(1);
     effect._02.set(0);
     effect._04.set((short)0);
-    manager._10.colour_1c.set(255, 0, 0);
+    if(Config.changeAdditionOverlayRGB()) {
```
LordMonoxide, line 887: If you invert this `if` statement you can merge the default colour hack with an `||` so you don't need two equivalent branches

src/main/java/legend/game/combat/SEffe.java
```diff
@@ -222,9 +227,11 @@ private SEffe() { }
 
   private static final Value _800fb7bc = MEMORY.ref(1, 0x800fb7bcL);
 
+  /** Some kind of mysterious global 2-hit addition array */
   private static final Value _800fb7c0 = MEMORY.ref(1, 0x800fb7c0L);
 
-  private static final Value _800fb7f0 = MEMORY.ref(1, 0x800fb7f0L);
+  /** Fours sets of color values used for addition overlay borders; only last actually used */
```
LordMonoxide, line 233:
```suggestion
  /** Four sets of colour values used for addition overlay borders; only last actually used */
```

src/main/java/legend/game/combat/SEffe.java
```diff
@@ -222,9 +227,11 @@ private SEffe() { }
...
+  public static final ArrayRef<UnsignedByteRef> additionBorderColors_800fb7f0 = MEMORY.ref(1, 0x800fb7f0L, ArrayRef.of(UnsignedByteRef.class, 0xc, 0x1, UnsignedByteRef::new));
```
LordMonoxide, line 234:
```suggestion
  public static final ArrayRef<UnsignedByteRef> additionBorderColours_800fb7f0 = MEMORY.ref(1, 0x800fb7f0L, ArrayRef.of(UnsignedByteRef.class, 0xc, 0x1, UnsignedByteRef::new));
```
We use the _proper_ spelling everywhere else

src/main/java/legend/game/modding/coremod/config/AdditionOverlayConfigEntry.java
```diff
@@ -0,0 +1,45 @@
...
+      return dropdown;
+    });
+  }
+
+  private static boolean validator(final AdditionOverlayMode val) {
+    return true;
+  }
+
+  private static byte[] serializer(final AdditionOverlayMode val) {
+    return new byte[]{(byte)val.ordinal()};
```
LordMonoxide, line 35:
```suggestion
    return new byte[] {(byte)val.ordinal()};
```

src/main/java/legend/game/modding/coremod/config/AutoAdditionConfigEntry.java
```diff
@@ -0,0 +1,44 @@
...
+      return dropdown;
+    });
+  }
+
+  private static boolean validator(final AutoAdditionMode val) {
+    return true;
+  }
+
+  private static byte[] serializer(final AutoAdditionMode val) {
+    return new byte[]{(byte)val.ordinal()};
```
LordMonoxide, line 34:
```suggestion
    return new byte[] {(byte)val.ordinal()};
```

src/main/java/legend/core/Config.java
```diff
@@ -82,11 +103,18 @@ public static void toggleSaveAnywhere() {
   }
 
   public static boolean autoAddition() {
+    if(gameState_800babc8 != null) {
```
LordMonoxide, line 106: This effectively makes the config file value unused, since the config registry entry has a default value anyway. It will only apply on the title screen before a game is loaded, where it doesn't have meaning anyway. I'm going to be extending the config system this release to support config values that are global so the Config class will be removed anyway.

Update where these methods are being called to use the game state directly and remove the auto-addition methods here. We can leave it in the debugger for now since we don't have a way to change options in combat yet.

LordMonoxide (changes requested):

src/main/java/legend/core/Config.java
```diff
@@ -183,6 +211,66 @@ public static void setBattleRGB(final int rgb) {
     properties.setProperty("battle_ui_colour_change", "true");
   }
 
+  public static int getAdditionOverlayRGB() {
```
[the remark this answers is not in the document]
theflyingzamboni: Should I change the others while I'm at it? I was copying the existing style in that file.

src/main/java/legend/game/combat/SEffe.java
theflyingzamboni: Don't tempt me to change *all* of the other spellings to be consistently correct. :stuck_out_tongue:

src/main/java/legend/core/Config.java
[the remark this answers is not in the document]
LordMonoxide: Yes you might as well

LordMonoxide (approved):

PR #478, opened 2023-05-07
Restore menu widths to original truncated size (closes #472)

### LINKED ISSUE #472: Black spacing on the right of battle exp ui
1. Is this retail?
2. Is this from the scaler?
3. It looks bad.
4. I vote to center this screen either way.

image

LordMonoxide:
Also makes it so display height can be adjusted

### DIFF
--- a/src/main/java/legend/game/SItem.java
+++ b/src/main/java/legend/game/SItem.java
@@ -74,7 +74,7 @@ import static legend.game.Scus94491BpeSegment.loadFile;
 import static legend.game.Scus94491BpeSegment.loadSupportOverlay;
 import static legend.game.Scus94491BpeSegment.memcpy;
 import static legend.game.Scus94491BpeSegment.scriptStartEffect;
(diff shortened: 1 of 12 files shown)


src/main/java/legend/game/SItem.java
```diff
@@ -2330,7 +2330,7 @@ public static void renderPostCombatReport() {
             FUN_8010d050(InventoryMenuState._18, 0x1L);
           } else {
             // Some items remaining
-            setWidthAndFlags(384);
+            resizeDisplay(368, 240);
```
Ink230, line 2333: Why not make life easier with enums wherever resizeDisplay is called.

Ink230 (approved):

src/main/java/legend/game/SItem.java
LordMonoxide: I thought about this for a while but I'm okay with this as-is. These will be changing again in the near future.

PR #491, opened 2023-05-12
Camera logging

theflyingzamboni:
Just in case we need it at some point. Turned off by default.


Ink230 (approved):
Is [CAMERA] needed in the logged string?

src/main/java/legend/game/combat/Bttl_800d.java
```diff
@@ -162,6 +166,9 @@
 public final class Bttl_800d {
   private Bttl_800d() { }
 
+  private static final Logger LOGGER = LogManager.getFormatterLogger(String.class);
```
LordMonoxide, line 169: Loggers must always use the current class as their name. You can omit the parameter entirely and it will automatically use the current class.

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #499, opened 2023-05-13
Fix submap transition crash (closes #498), fix submap texture stretch (closes #497), change default res to 1

### LINKED ISSUE #497: Hellena Prison Second Tower Door placement, wrong
Pretty much as title say, Hellena Prison second Tower door (where Shana is captive) seems to be in the wrong place...
Door_Problem

### LINKED ISSUE #498: crash trying to transition submaps after lowering resolution scale to 1
[res scale drop.txt](https://github.com/Legend-of-Dragoon-Modding/Legend-of-Dragoon-Java/files/11469615/res.scale.drop.txt)
[Doel.zip](https://github.com/Legend-of-Dragoon-Modding/Legend-of-Dragoon-Java/files/11469640/Doel.zip)
[res test.zip](https://github.com/Legend-of-Dragoon-Modding/Legend-of-Dragoon-Java/files/11469643/res.test.zip)

This is starting at the save point in the Black Castle. If I go into the options menu and lower the resolution scale to 1, I get the above crash when I try to transition to the next screen with Kongol on the ramp. The crash happens even if I lower the res scale to 1, and then load back in using that save(res test save file). I do not experience this crash if I leave the resolution scale at 2.

This is on the devbuild updated May 12th using a save originally made on the May 1st devbuild.

LordMonoxide:


Ink230 (approved):
Ah, 368 vs 384, the age-old debate that has kept philosophers up at night for centuries.

Let's start with 368. This number has some serious swagger. It's the kind of number that walks into a room and immediately commands attention. It's got a bit of a bad-boy vibe, like it's always up to something sneaky and mischievous. If 368 were a person, it would probably have a leather jacket and a motorcycle.

On the other hand, we have 384. This number is a bit more buttoned-up. It's the kind of number that always follows the rules and color-codes its sock drawer. It's a bit of a nerd, but in a lovable way. If 384 were a person, it would probably have a pocket protector and a calculator.

So, which one is better? Well, that depends on what you're looking for. If you want a number that's going to shake things up and keep you on your toes, then 368 is your guy. But if you're looking for a number that's reliable and trustworthy, then 384 is the way to go. In either case you're still dealing with LoD code.

Of course, at the end of the day, both of these numbers are just that: numbers.

PR #502, opened 2023-05-13
Removes transparency command from MCQs

theflyingzamboni:
MCQs use gp0 command 7c, which does not have transparency.


LordMonoxide:
One thing, can you verify that WMAP's MCQ renderer is correct? It uses B+F (see 800e4b14)

theflyingzamboni:
Of course there are two MCQ renderers. -__-

LordMonoxide (approved):

PR #508, opened 2023-05-15
Perfect button mashing on keyboard and controller (closes #412)

### LINKED ISSUE #412: Item mashing is broken
SC seems not to be registering all inputs when mashing. This seems to become more of an issue when you mash faster. Swiping quickly, I cannot get the percentage above 110%, but mashing (which is slower) can get me over 150% (though this is still low).

LordMonoxide:


src/main/java/legend/core/opengl/Window.java
```diff
@@ -255,20 +262,21 @@ public float getScale() {
...
+      Action nextAction = null;
 
-      glfwSwapBuffers(this.window);
+      for(final Action action : this.actions) {
+        action.tick();
 
-      if(this.fpsLimit != Integer.MAX_VALUE) {
-        while(System.nanoTime() <= timer) {
-          DebugHelper.sleep(0);
+        if(nextAction == null || action.nanosUntilNextRun() < nextAction.nanosUntilNextRun()) {
```
Ink230, line 273: Would using a queue (priorityqueue in Java seems like the right one for this) instead?

I think you could prioritize actions based on which one has the smallest time left to run, and run those first.

This way they can be executed in the correct order based on nanosUntilNextRun. Instead of in the order they were applied to the list?

?Technically? some actions could be executed late if the list is long enough / an action is blocking enough?
?Technically? not doing the one with the least nanosUntilNextRun would cascade errors through new assigned timings?

src/main/java/legend/game/input/Controller.java
Ink230, line 1: Rumble handling?

Ink230 (approved):
Moderately fantastic PR. ✔

src/main/java/legend/core/opengl/Window.java
LordMonoxide: Hmm, I just looked into priorityqueue and it would get a bit weird... this will only sleep until the next task is ready so in practice it'll work the way you're describing

src/main/java/legend/game/input/Controller.java
LordMonoxide: Soon™️

LordMonoxide:
When a new controller is plugged in and no controller was plugged in previously, automatically switch to that controller

PR #511, opened 2023-05-16
Fixes heal particle movement

theflyingzamboni:
a2._50 is an SVEC, not a VEC, so param 2 + 54 in ghidra equates to the Z value rather than the Y value.


Ink230 (approved):
X, Y, and Z were triplets...but one of them was different from the others. You see, the three came from a family of SVECTORS but Y could never quite get their sin and cos to meet the dimensional expectations from parents U and V. 

Y thought at first it was their product getting all crossed, but to no avail. But one day while pondering if they'd have find out why they were so different, Y saw something out on the planes.

Curious, Y transformed across to the stranger. As Y got larger, they began to feel the fear creeping upon them about the stranger. After it was too late to identify back, Y saw the rumour as fact before them. There, poised and ready stood the Reboot adventurer who had been cleansing all coordinates in the mainframe.

Before Y could project its current predicament, the Reboot adventurer wiped Y out of existence before muttering, "Pesky VECTOR".

LordMonoxide (approved):

PR #538, opened 2023-05-20
Fix encounter rates (closes #537)

### LINKED ISSUE #537: Update Encounter Rate For "None" mode on World Map
image
Map encounter rate should be 0

LordMonoxide:


theflyingzamboni (approved):
Those look like some nice numbers.

PR #541, opened 2023-05-20
Convert a bunch of configs (closes #531)

### LINKED ISSUE #531: Convert remaining configs to new config system

LordMonoxide:


theflyingzamboni (changes requested):
For the settings that you moved from the debugger, could you give them a Refresh button like I did for auto-addition so that the debugger values can be re-synced? Actually, just a single refresh button for all now-options-controlled settings would be better now.

theflyingzamboni:
Also, would you mind changing regular auto-addition to a checkbox now that you've made the element?

theflyingzamboni (approved):

PR #559, opened 2023-05-23
Fix sounds hanging (closes #208)

### LINKED ISSUE #208: Battle Transition SFX hangs for long periods
Although we can use the delete key to cut a hanging sound effect short, it becomes laborious after many repeated uses per session. If the hang cannot be fixed right now, recommend muting or removing that sound effect.

LordMonoxide:


Ink230 (approved):
Inspected, Accepted.

PR #562, opened 2023-05-24
Adds virtualSize to FileData

### LINKED ISSUE #558: Lenus 2 attack issues
First issue is on the attack where she throws her weapon up above the party and then attacks them underwater. This attack seems to softlock if there is a Dragoon in the battle. It has softlocked on a Shana special, a Meru special, and just Shana being dragoon with Dart/Rose in human form. This attack works fine if all 3 members are in human form.

Second issue is the attack where she creates the 4 pillars that fill up with water. This attack is not hitting everyone in the party when someone is in dragoon form (special or not). When all 3 members are human, attack works fine.

[Lenus 2.zip](https://github.com/Legend-of-Dragoon-Modding/Legend-of-Dragoon-Java/files/11539166/Lenus.2.zip)

theflyingzamboni:
It turns out that sometimes virtual files with sizes are referenced by indices hardcoded into scripts, and not having them present causes a softlock when the animation cannot be found (as with Lenus' water attacks and the floating animation). I double-checked the retail code in ghidra, and all the places we do `real()` checks, retail is just checking for `size != 0`, regardless of whether the file is a duplicate. To match that, I added a virtualSize to FileData that always holds the size found in the MRG table, and `real()` now checks that this isn't zero.

Fixes #558

### DIFF
--- a/src/main/java/legend/game/unpacker/Unpacker.java
+++ b/src/main/java/legend/game/unpacker/Unpacker.java
@@ -39,6 +39,8 @@ import java.util.stream.StreamSupport;
 import static legend.game.Scus94491BpeSegment.getCharacterName;
 
 public final class Unpacker {
+  private static final Pattern MRG_ENTRY = Pattern.compile("[=;]");
+
   private Unpacker() { }
 
   private static final String[] DISK_IDS = {"SCUS94491", "SCUS94584", "SCUS94585", "SCUS94586"};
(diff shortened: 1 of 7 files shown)


Ink230 (approved):
No silly issues, though it is fun to see the same parameter passed twice. Let Monoxide review.

src/main/java/legend/game/unpacker/Unpacker.java
```diff
@@ -138,17 +138,21 @@ public static List<FileData> loadDirectory(final String name) {
       if(Files.exists(mrg)) {
         try(final BufferedReader reader = Files.newBufferedReader(mrg)) {
           final Int2IntMap fileMap = new Int2IntArrayMap();
+          final ArrayList<Integer> fileSizes = new ArrayList<>();
```
LordMonoxide, line 141: Make this an Int2IntMap like the previous one, keyed by virtual ID, having it as a list seems brittle if we end up having to change this code

PR #587, opened 2023-05-31
chore: build script updates for reasons

zack37:


LordMonoxide:
Thanks, had to modify this a bit to work with our cross-platform build scripts but we now support ARM on all platforms and ARM devbuilds will be auto-generated as well

PR #640, opened 2023-06-19
Finished dememulating weapon trail effect

theflyingzamboni:
Kanzas didn't explode in testing, so it should be working.


Ink230 (dismissed):
Make .set and .get great again.

LordMonoxide (approved):

PR #648, opened 2023-06-26
Fixes crash when using weapons that apply poison

theflyingzamboni:
Prior to stats changes, the fields grouped into SpellStats0c were set to 0 when spellID was >127. This got changed to setting bobj.spell_94 to null, which throws an error when the checkHit method tries to access one of the values. Thought about adding a null check, but thought replicating the prior behavior with a default constructor would be safer.

### DIFF
--- a/src/main/java/legend/game/types/SpellStats0c.java
+++ b/src/main/java/legend/game/types/SpellStats0c.java
@@ -1,7 +1,7 @@
 package legend.game.types;
 
 import legend.game.characters.Element;
-import legend.game.modding.coremod.elements.NoElement;
+import legend.game.modding.coremod.CoreMod;
 import legend.game.unpacker.FileData;
 
 public class SpellStats0c {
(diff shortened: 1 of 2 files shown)


src/main/java/legend/game/types/SpellStats0c.java
```diff
@@ -47,6 +47,23 @@ public static SpellStats0c fromFile(final String name, final String combatDescri
...
+    this.combatDescription = "";
+    this.targetType_00 = 0;
+    this.flags_01 = 0;
+    this.specialEffect_02 = 0;
+    this.damageMultiplier_03 = 0;
+    this.multi_04 = 0;
+    this.accuracy_05 = 0;
+    this.mp_06 = 0;
+    this.statusChance_07 = 0;
+    this.element_08 = Element.fromFlag(0);
```
LordMonoxide, line 61: Change to the actual element ref, Element.fromFlag won't work with mods and will be removed once we figure out how to safely remove it

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #658, opened 2023-07-03
Hell Mode additional changes

Zychronix:


Ink230 (commented):
1. Didn't crash when I loaded it. ✔
2. Shiranda causing problems as always. ✔
3. Character/Dragoon stats being 20 different arrays probably isn't a good idea and now seems like there is no blocker to make it proper with a nice OOP design.
4. Characters should at least be an enum (that can be added to by a mod event) and this enum would be enumareted for things like loading stats/exp/making init adjustments.. Then we won't ever be able to pretend Miranda doesn't exist. ✔

_Death to static._

PR #688, opened 2023-08-05
More Modding Events

Zychronix:
Character stats are now in the new CoreMod CharacterData Partially dememulated additions
Addition max level/additions to level change
Character max level change
New battle events
Event to bypass divine dragoon automatically setting SP Removed sp acquired cap


Ink230 (dismissed):
Mods. ✔

Ink230 (dismissed):
LGTM! ✔

PR #703, opened 2023-08-28
Script meta 7

LordMonoxide:


src/main/java/legend/game/SMap.java
```diff
@@ -8254,6 +8282,17 @@ public static FlowControl FUN_800f1060(final RunningScript<?> script) {
     return FlowControl.CONTINUE;
   }
 
+  @ScriptDescription("Unknown")
+  @ScriptParam(direction = ScriptParam.Direction.IN, type = ScriptParam.Type.INT, name = "p0")
+  @ScriptParam(direction = ScriptParam.Direction.IN, type = ScriptParam.Type.INT, name = "p1")
+  @ScriptParam(direction = ScriptParam.Direction.IN, type = ScriptParam.Type.INT, name = "p2")
```
LordMonoxide, line 8288: This is technically BOTH, but I don't think it's really intended to be used in that way. Can't decide if I should doc it as such or not

src/main/java/legend/game/SMap.java
```diff
@@ -8474,6 +8540,11 @@ public static FlowControl FUN_800f1b64(final RunningScript<?> script) {
     return FlowControl.CONTINUE;
   }
 
+  @ScriptDescription("Unknown, adds a triangle indicator (possible for a door)")
```
LordMonoxide, line 8543: possibly

src/main/java/legend/game/combat/Bttl_800c.java
```diff
@@ -4459,33 +4530,40 @@ public static int FUN_800cfc20(final Vector3f managerRotation, final VECTOR mana
...
-    script.params_20[3].set(sp0x40.getY());
-    script.params_20[4].set(sp0x40.getZ());
+    script.params_20[2].set(lw.transfer.getX());
+    script.params_20[3].set(lw.transfer.getY());
+    script.params_20[4].set(lw.transfer.getZ());
     return FlowControl.CONTINUE;
   }
 
+  @ScriptDescription("Gets a bounding box dimension for a battle entity")
+  @ScriptParam(direction = ScriptParam.Direction.OUT, type = ScriptParam.Type.INT, name = "size", description = "The size")
```
LordMonoxide, line 4562: Params are wrong

src/main/java/legend/game/combat/Bttl_800d.java
```diff
@@ -1213,6 +1257,9 @@ public static void FUN_800d3bb8(final ScriptState<AdditionNameTextEffect1c> stat
     //LAB_800d3d1c
   }
 
+  @ScriptDescription("Allocates an addition addition name display script state")
```
LordMonoxide, line 1260: addition addition

src/main/java/legend/game/combat/SEffe.java
```diff
@@ -4082,12 +4139,15 @@ public static FlowControl allocateAdditionOverlaysEffect(final RunningScript<? e
   }
 
   /**
-   * Script subfunc related to pausing ticking/rendering of addition overlay during counterattacks. v1 == 0 occurs when
+   * Script subfunc related to pause ticking/rendering of addition overlay during counterattacks. additionContinuationState == 0 occurs when
```
LordMonoxide, line 4142: pausing

src/main/java/legend/game/combat/SEffe.java
```diff
@@ -4533,6 +4601,8 @@ public static FlowControl allocateDragoonAdditionScript(final RunningScript<?> s
     return FlowControl.CONTINUE;
   }
 
+  @ScriptDescription("Gets the number of dragoon addition hits completed")
+  @ScriptParam(direction = ScriptParam.Direction.OUT, type = ScriptParam.Type.INT, name = "count", description = "The number of hits completed")
```
LordMonoxide, line 4605: Missing unused param

src/main/java/legend/game/SMap.java
theflyingzamboni: I think you're correct that the intent is just to modify the param for use in the method itself. The issue is, I can't find a use of this method anywhere right now, so I can't see whether the params are hardcoded values or not. If they are, then it should definitely be IN, but if it could be a storage param or something, then it would still have an OUT side effect. Given it's not used as far as I can find, maybe leave it IN but include a note?

PR #786, opened 2023-11-02
Zoom and location prompt overlays

### LINKED ISSUE #778: [main.opengl] Off-centered text on World Map pop-up screen.
Definitely for Bale, looks like it may be an issue for other areas as well. image
image

theflyingzamboni:
- Implemented the zoom overlay
- Refactored the wmap prompt popup into its own class
  - Plan to refactor highlight stuff as well in future when I refactor the Coolon travel prompt
- Fixed location prompt text alignment issue

Closes #778, #784, #785


src/main/java/legend/game/wmap/WmapPromptPopup.java
```diff
@@ -0,0 +1,192 @@
...
+    PROMPT,
+    OPTIONS,
+    ALT_TEXT,
+    THUMBNAIL
+  }
+
+  private final MV transforms = new MV();
+
+  private TextObj prompt;
+  final private Vector3f promptTranslation = new Vector3f();
```
LordMonoxide, line 27: Flip these to `private final`

src/main/java/legend/game/wmap/WmapPromptPopup.java
```diff
@@ -0,0 +1,192 @@
...
+      this.transforms.transfer.set(this.promptTranslation);
+      RENDERER.queueOrthoOverlayModel(this.prompt, this.transforms);
+    }
+
+    if(!this.options.isEmpty()) {
+      TextObj option;
+      for(int i = 0; i < this.options.size(); i++) {
+        option = this.options.get(i);
+        if(option != null) {
+          this.transforms.transfer.set(this.optionsTranslation);
```
LordMonoxide, line 143: Not strictly necessary, but you could set the translation once before the loop, and then just add `this.optionSpacing` each iteration (same below)

src/main/java/legend/game/wmap/WmapPromptPopup.java
```diff
@@ -0,0 +1,192 @@
...
+    if(this.prompt != null) {
+      this.prompt.delete();
+      this.prompt = null;
+    }
+
+    if(!this.options.isEmpty()) {
+      for(int i = 0; i < this.options.size(); i++) {
+        if(this.options.get(i) != null) {
+          this.options.get(i).delete();
+          this.options.set(i, null);
```
LordMonoxide, line 173: Rather than set null, after the loop just clear the list (same below)

PR #787, opened 2023-11-04
Tracking for Objs

### LINKED ISSUE #788: Implement Obj tracking

theflyingzamboni:
This doesn't feel right, but it sure makes it easy to find the issue.

Closes #788

### DIFF
--- a/src/main/java/legend/core/opengl/Obj.java
+++ b/src/main/java/legend/core/opengl/Obj.java
@@ -33,7 +33,7 @@ public abstract class Obj {
   public static void clearObjList() {
     for(int i = objList.size() - 1; i >= 0; i--) {
       if(shouldLog) {
-        LOGGER.info("Leaked: %s", objList.get(i).name);
+        LOGGER.warn("Leaked: %s", objList.get(i).name);
       }
       objList.get(i).delete();
     }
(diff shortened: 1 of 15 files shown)


src/main/java/legend/core/opengl/Obj.java
```diff
@@ -1,11 +1,33 @@
...
+import java.util.List;
 
-public interface Obj {
-  boolean shouldRender(@Nullable final Translucency translucency);
-  void render(@Nullable final Translucency translucency);
-  void delete();
+public abstract class Obj {
+  private static final Logger LOGGER = LogManager.getFormatterLogger();
+
+  protected static final List<Obj> objList = new ArrayList<>();
```
LordMonoxide, line 14: Make this private and implement the `delete` method on this class to remove the `Obj` from the list. Then update the subclass `delete`s to call `super.delete()`

src/main/java/legend/core/opengl/Obj.java
```diff
@@ -1,11 +1,33 @@
...
+  private final String name;
+
+  public Obj(final String name) {
+    this.name = name;
+    objList.add(this);
+  }
+
+  public static void reportLeaks() {
+    if(!Obj.objList.isEmpty()) {
```
LordMonoxide, line 23: 1. You don't need the isEmpty if statement
2. Might as well delete all undeleted objects and clear the list

LordMonoxide (approved):

LordMonoxide (approved):

PR #845, opened 2023-11-13
Implemented wmap smoke effect in opengl

theflyingzamboni:
Please double check my condition condensing.

Transparency obviously does not look right currently.


src/main/java/legend/game/wmap/WMap.java
```diff
@@ -6284,7 +6284,6 @@ private void handleSmokeEffect() {
   @Method(0x800edbc0L)
   private void renderSmoke() {
     final Vector3f rotation = new Vector3f(); // Just (0, 0, 0)
```
LordMonoxide, line 6286: Dunno why I left this in in the first place, but just remove this and the rotation call, since it does nothing

LordMonoxide (approved):

PR #899, opened 2023-12-03
Refactored zoomState_1f8 to use ZoomState enum

theflyingzamboni:
Also yeeted two of the zoom states as they seemed to be essentially unnecessary, and I have not noticed any issue using the map zoom with this refactor. Finally, I caught a couple minor tangential decomp and 60fps bugs related to updating lights when zooming.

### DIFF
--- a/src/main/java/legend/game/wmap/WMapModelAndAnimData258.java
+++ b/src/main/java/legend/game/wmap/WMapModelAndAnimData258.java
@@ -33,7 +33,6 @@ public class WMapModelAndAnimData258 {
     LOCAL,
     CONTINENT,
     TRANSITION_MODEL_OUT,
-    TRANSITION_ARROW_SIZE,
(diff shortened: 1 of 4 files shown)


src/main/java/legend/game/wmap/WMapModelAndAnimData258.java
```diff
@@ -28,18 +28,13 @@ public enum MapTransitionDestinationType {
     WORLD_MAP
   }
 
-  /**
-   * TODO Look more into zoom code before refactoring to use states, might be able
-   *  to reduce number.
-   */
+
   public enum ZoomState {
```
LordMonoxide, line 32: Just make a note of which retail values map to which enums

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #957, opened 2023-12-20
Implemented SMAP snow effect

### LINKED ISSUE #923: Implement SMAP snow
Uses the same struct as smoke.

theflyingzamboni:
- Named fields and vars
- Refactored code into own class
- Converted to floats where needed
- Implemented modern RNG to improve snow distribution
- Used shadow texture instead of plain quad for snow obj

Closes #923


src/main/java/legend/game/submap/SMap.java
```diff
@@ -7401,40 +7163,29 @@ private FlowControl scriptReinitializeSmokePlumeForIntermittentBursts(final Runn
...
+  private FlowControl scriptInitSnowParticleData(final RunningScript<?> script) {
+    this.snowState_800f9e60 = script.params_20[0].get();
 
+    //LAB_800f21d0
+    if(this.snowState_800f9e60 == 1) {
       if(script.params_20[2].get() < 0) {
         script.params_20[2].neg();
       }
 
+      final float stepAngleMax = script.params_20[4].get() / 4096.0f * MathHelper.TWO_PI;
```
LordMonoxide, line 7182: MathHelper.psxDegToRad

src/main/java/legend/game/submap/SnowEffect.java
```diff
@@ -0,0 +1,241 @@
...
+    public float brightness_34;
+
+    // public SnowParticleInstance3c next_38;
+  }
+
+  public SnowEffect(final float stepAngleMax, final int translationScaleX, final int stepXMax, final int stepYDivisor) {
+    this.particleData = new SnowParticleData18(stepAngleMax, translationScaleX, stepXMax, stepYDivisor);
+    this.initSnowEffect();
+
+    this.particle = new QuadBuilder("Snowflake")
```
LordMonoxide, line 72: Having anything that communicates with the GPU in a constructor can make it tricky to instantiate the object, because it has to be done in the main thread. If we were to refactor SMAP to load in a secondary thread, for example, this would crash. That's why I usually build Objs in the render methods themselves, if they haven't been built yet, since we're guaranteed to be on the render thread there. It's a little messy, but it gets the job done.

src/main/java/legend/game/submap/SnowEffect.java
```diff
@@ -0,0 +1,241 @@
...
+      } else {
+        //LAB_800ee750
+        inst.angleStep_0c = this.rand.nextFloat(this.particleData.stepAngleMax_08);
+      }
+
+      this.particles[i] = inst;
+
+      //LAB_800ee770
+      this.snowEffectTick = this.snowEffectTick + 1 & 0xff;
+      this.snowOffsetXTick = this.snowOffsetXTick + 1 & 0xf;
```
LordMonoxide, line 172: Are these good for variable frame rates?

src/main/java/legend/game/submap/SnowEffect.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    //LAB_800ee988
+    this.snowEffectTick++;
+    if(this.snowEffectTick >= 6) {
+      this.snowEffectTick = 0;
+    }
+
+    //LAB_800ee9b4
+    this.snowOffsetXTick = this.snowOffsetXTick + 1 & 0xf;
```
LordMonoxide, line 233: Same here, at the if statement above

LordMonoxide (changes requested):

src/main/java/legend/game/submap/SMap.java
theflyingzamboni: Derp

src/main/java/legend/game/submap/SnowEffect.java
theflyingzamboni: These variables only increment per particle processed, not per frame, and are only used when initializing a new particle traversal (either init or wrap).

src/main/java/legend/game/submap/SnowEffect.java
theflyingzamboni: See above

LordMonoxide (approved):

PR #1373, opened 2024-05-29
Input Toggle for GameSpeedMultiplier

Ink230:
## Motivation
Needing to turn the game speed multiplier back to 1 (off) and then back to some higher game speed multipler. And then doing this a thousand times when testing.

## Scope
I made it so that the extra Config parameter is sourced from config.conf (on load) or Debugger (on set). The +/- shortcuts were intentionally ignored to give a new degree of freedom. Aka, use +/- to customize your session but use the toggle to revert to your preferred config state.

## Hot key
Control + Q.

Not sure what system keys are reserved for. Control+Q is easily accessible (one-handed shortcuts btw) and not in use elsewhere.

## Future
There's a lot more that can be done with game speed multipliers. At the moment, it is all tied to the gameSpeedMulitplier static variable in Config. This value should probably be managed in a class that can then pull from Config and manage the state of the gameSpeedMultiplier option. Ex: Customizations for the multipler specific to submaps, combat, hold toggles, holdandrelease toggles, multiple hotkeys/rebinds for different multiplier stages, changing the base of the onandoff toggle...etc.

Don't think my PR merits this above overhead.

### DIFF
--- a/src/main/java/legend/game/Scus94491BpeSegment.java
+++ b/src/main/java/legend/game/Scus94491BpeSegment.java
@@ -163,6 +163,7 @@ import static legend.game.Scus94491BpeSegment_800c.sequenceData_800c4ac8;
 import static legend.game.combat.environment.StageData.stageData_80109a98;
 import static org.lwjgl.glfw.GLFW.GLFW_KEY_DELETE;
 import static org.lwjgl.glfw.GLFW.GLFW_KEY_F12;
+import static org.lwjgl.glfw.GLFW.GLFW_KEY_Q;
 import static org.lwjgl.glfw.GLFW.GLFW_KEY_W;
 import static org.lwjgl.glfw.GLFW.GLFW_MOD_CONTROL;
 
@@ -373,6 +374,14 @@ public final class Scus94491BpeSegment {
 
       if((mods & GLFW_MOD_CONTROL) != 0 && key == GLFW_KEY_W && currentEngineState_8004dd04 instanceof final Battle battle) {
         battle.endBattle();
+      }
+
+      if((mods & GLFW_MOD_CONTROL) != 0 && key == GLFW_KEY_Q) {
+        if(Config.getGameSpeedMultiplier() == 1) {
+          Config.setGameSpeedMultiplier(Config.getLoadedGameSpeedMultiplier());
+        } else {
+          Config.setGameSpeedMultiplier(1);
+        }
       }
     });
 
--- a/src/main/java/legend/core/Config.java
+++ b/src/main/java/legend/core/Config.java
@@ -75,6 +75,7 @@ public final class Config {
   }
 
   private static int gameSpeedMultiplier = 1;
+  private static int loadedGameSpeedMultiplier = 1;
 
   public static boolean lowMemoryUnpacker() {
     return readBool("low_memory_unpacker", false);
@@ -135,6 +136,14 @@ public final class Config {
   public static void setGameSpeedMultiplier(final int multiplier) {
     gameSpeedMultiplier = multiplier;
     properties.setProperty("game_speed_multiplier", String.valueOf(multiplier));
+  }
+
+  public static int getLoadedGameSpeedMultiplier() {
+    return loadedGameSpeedMultiplier;
+  }
+
+  public static void setLoadedGameSpeedMultiplier(final int multiplier) {
+    loadedGameSpeedMultiplier = multiplier;
   }
 
   public static Vector3f getBattleRgb() {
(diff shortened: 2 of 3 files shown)


src/main/java/legend/game/Scus94491BpeSegment.java
```diff
@@ -374,6 +375,14 @@ public static void gameLoop() {
       if((mods & GLFW_MOD_CONTROL) != 0 && key == GLFW_KEY_W && currentEngineState_8004dd04 instanceof final Battle battle) {
         battle.endBattle();
       }
+
+      if((mods & GLFW_MOD_CONTROL) != 0 && key == GLFW_KEY_Q) {
+        if (Config.getGameSpeedMultiplier() == 1) {
```
LordMonoxide, line 380: Formatting

LordMonoxide, line 380: What happens if they toggle it with ctrl+Q and then adjust the modifier back to 1 with =/-?

Wouldn't it make more sense to have a boolean "disable multiplier" or something like that?

Ink230: **Pretext**
1. Basically, we are asking what the purpose of = and - are for. Are they to change what is currently being rendered to your eyes (always on) or are they to change a setting you may or may not be experiencing at the moment (disable multiplier).

2. This feature doesn't exclude a 'disable multiplier' option from existing. Both have merit to exist together.

**'always on'**
- GameSpeedMultiplier: 8
- Control + Q => GameSpeedMultiplier: 1
- = => GameSpeedMultiplier: 2
- Control + Q => GameSpeedMultiplier: 1
- Control + Q => GameSpeedMultiplier: 2

In the implementation, =/- are always only changing the _current value of how fast your game is CURRENTLY rendering... via GameSpeedMultiplier_. This is because, there is only ever one variable controlling your game speed, GameSpeedMultiplier. There's no other states or variables to keep track of. So what you see is what you get.

I like this behaviour as its simple and you can always see what the state of GameSpeedMultiplier is. If it is 1, you are rendering at x1 speed. If it is 8, you will be rendering at x8 speed. There's no layers of state for you to worry about, just press buttons to see exactly what you're in.

The niche behaviour of setting it to 1 and having the option to control it from there with =/- is also useful for a niche of use cases.

This is effectively also just a 'reset gamespeedmultiplier' with the caveat that it remembers your last used value...to make it toggle-like.

**'disable multiplier'**
- GameSpeedMultiplier: 8
- Control + Q => GameSpeedMultiplier: 8
- -x3 => GameSpeedMultiplier: 5 (but you don't see this)
- Control + Q => GameSpeedMultiplier: 5 (you toggled off at 8, and resume at 5)

As stated, I don't like that you can change the GameSpeedMultiplier to something you're not experiencing, for something like rendering. I can press =/- and see no effect or change to how fast I'm rendering the game. I have to hit Control + Q in this scenario to see what values I ended up on and then feel if that is good enough. The behaviour is disconnected.

We can add more UI tooling to inform the user of the current state of course.

Of course, very easy to convert this PR to this implementation. Do not mind doing that, but I will note that I see use cases for both types of settings.

**niche use case example**
Overworld/Submaps are 8x. I spend most of my time in these parts of the game. My time spent here is long and often. However, Battle occurs sparingly but I don't want to use 8x. It's just too fast.

I want to use x1 to x3. x1 when I'm doing additions. x2/x3 when others are attacking (or defending) and I just want to see damage numbers / heal numbers. 

With this Control + Q and =/- combination, I can meet this flow quite nicely and be flexible in the game speed for the frequency I need it to be flexible for across the different game contexts. _This is my original inspiration for the current implementation._ Too much Irongoon testing.

Of course, we can add more settings to make this more convenient, instead of tentpoling on the current PR. But I think the current PR is fine enough.

**'third option(s)'**
- Disable =/- when control+q is active
- Have =/- only affect the current session and not what is in the config file or debugger, control + q then only refers to current session values OR config file/debugger values
- Be able to set default game speeds for Overworld, Submap, Combat. Control + Q toggles them to 1/disables them/toggles them. =/- role is to then do any of the above.

I haven't played with these to know if they would be useful or a hindrance or just neutral.

**In the end,**
I think each way is equivalent for 99.5% of use cases. People are only going to _use these keys mutually exclusively_. Control + Q as a toggle, or =/- to dial in a setting...rarely will people be intermingling these key presses.

I agree the 'disable multiplier' seems more standard, but from doing Ironmon on various Gen 1-5 emulators, having a setting that inc/dec the game speed from the current game speed setting is a nice to have. Being able to toggle different game speeds and then adjusting those on the fly is also nice. These emulators of course have both options, both the current style of implementation and a separate 'disable all speedups'.

Further, if there is a need to change the PR or switch it to disable multiplier or if the whole inputs for keyboard inputs is revamped...this is a fairly trivial change to make or to remove. 

**future**
For the current implementation, the feature set can be expanded to
- Toggle to some other default speed other than 1 (customizable by user)
- Hit multiple speed breakpoints as you toggle through them (1 -> 4 -> 6 -> 2 -> 9 -> 3)

Ink230: updated

LordMonoxide (approved):

PR #1389, opened 2024-05-31
Combat Hot Keys

Ink230:
### Motivation
- Playing at high game speed multipliers makes navigating the Battle menu difficult
- Supports accessibility
  - Makes one-handed playing easier
  - Permits controller + keyboard usages for accessibility (rebinds)
- The game inputs are live but their inputs are unmapped to any actions
  - Since the its using the integrated InputAction system, we can scope these to only combat contexts
  - Will be seamless with any future hotkey/input rebind functionality later

### Implementation
- Sounds are played on invalid selection w/ a validation check on the hotkey
  - Also a utility method within BattleHud that could replace a few dozen playSound lines (refactoring some logic however)
- battleMenu_800c6c34
  - Added 2 utility methods to retrieve icon status (todo: remove all the duplicated original utils checking for these icon behaviours)
-  iconFlags_800c7194 (BattleHud)
  - Uses this for indices in a non-human readable format to not have extra random ints in BattleHud
  - iconFlags_800c7194 = {4, 1, 5, 6, 2, 9, 3, 7} for reference, enum where btw
- You can not take any action otherwise not enabled for the battle context (seamless integration by big brain Ink)

### Gameplay Changes
- Added a 'Do nothing' bind (skip turn)
  - Useful for regular testing where you just do not want to take any turn action
  - Also useful while in Dragoon form to take no action
    - WTB 'Run' icon being enabled in Dragoon form RFC

### Mappings
- Attack -> None, since the default is already X on turn start (unnecessary)
- Guard -> 1 -> 1 -> BUTTON_SHOULDER_RIGHT -> Intentional: to be on the same left hand side as the movement joystick
- Item/DMagic -> 3, 5 -> 2, 6 -> BUTTON_WEST -> Just made sense, if BUTTON_NORTH is Addition menu + most modern games would put a 'open menu' button on the ordinal buttons alongside Select, Cancel
- Skip Turn -> Use a disabled icon and set the action to that int (nice hack ✔)
  - If Items can be used in Dragoon form this poses a problem...until then!
- Dragoon -> 2 -> 4 -> Joystick Right UP -> To POWER UP, accessible on right hand where you take action
- Special -> 7 -> 7 -> Joystick Right DOWN -> Needs to be opposite to Dragoon to avoid accidents on the Joystick

Of course, better suggestions for mappings are appreciated. I think these largely make the most sense for usability/flow.

### Challenges
- battleMenu_800c6c34.iconFlags_10 changes its values based on battle context
 
### Future
- I noticed 8 is missing as an icon flag
- De-Dragoon icon possibilities...seems natural
- Would be the only icon I personally think would merit as a core game play extension to the default iconFlags and to take advantage of
- Unless 8 is used somewhere and I'm ignorant

### DIFF
--- a/src/main/java/legend/game/combat/ui/BattleMenuStruct58.java
+++ b/src/main/java/legend/game/combat/ui/BattleMenuStruct58.java
@@ -14,6 +14,8 @@ import legend.game.types.Translucency;
 
 import javax.annotation.Nullable;
 
+import java.util.Arrays;
+
 import static legend.game.combat.ui.BattleHud.battleMenuIconHeights_800fb6bc;
 
 public class BattleMenuStruct58 {
@@ -340,5 +342,17 @@ public class BattleMenuStruct58 {
         this.targetArrows[i] = null;
       }
     }
+  }
+
+  public boolean isIconEnabled(int value) {
+    return Arrays.stream(this.iconFlags_10, 0, this.iconFlags_10.length)
+      .anyMatch(i -> i == value);
+  }
+
+  public int retrieveIconEnabled(int... values) {
+    return Arrays.stream(this.iconFlags_10, 0, this.iconFlags_10.length)
+      .filter(i -> Arrays.stream(values).anyMatch(value -> value == i))
+      .findFirst()
+      .orElse(0);
   }
 }
--- a/src/main/java/legend/game/combat/ui/BattleMenuStruct58.java
+++ b/src/main/java/legend/game/combat/ui/BattleMenuStruct58.java
@@ -344,12 +344,12 @@ public class BattleMenuStruct58 {
     }
   }
 
-  public boolean isIconEnabled(int value) {
+  public boolean isIconEnabled(final int value) {
     return Arrays.stream(this.iconFlags_10, 0, this.iconFlags_10.length)
       .anyMatch(i -> i == value);
   }
 
-  public int retrieveIconEnabled(int... values) {
+  public int retrieveIconEnabled(final int... values) {
     return Arrays.stream(this.iconFlags_10, 0, this.iconFlags_10.length)
       .filter(i -> Arrays.stream(values).anyMatch(value -> value == i))
       .findFirst()
--- a/src/main/java/legend/game/combat/ui/BattleHud.java
+++ b/src/main/java/legend/game/combat/ui/BattleHud.java
@@ -2365,31 +2365,31 @@ public class BattleHud {
           }
         }
 
+        // Dragoon
         if(Input.pressedThisFrame((InputAction.JOYSTICK_RIGHT_BUTTON_UP))) {
           selectedAction = this.battleMenu_800c6c34.isIconEnabled(iconFlags_800c7194[4]) ? iconFlags_800c7194[4] : 0;
           this.checkInvalidSelectedAction(selectedAction);
         }
 
+        // Special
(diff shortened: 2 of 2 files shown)


src/main/java/legend/game/combat/ui/BattleMenuStruct58.java
```diff
@@ -341,4 +343,16 @@ public void delete() {
       }
     }
   }
+
+  public boolean isIconEnabled(final int value) {
+    return Arrays.stream(this.iconFlags_10, 0, this.iconFlags_10.length)
+      .anyMatch(i -> i == value);
+  }
+
+  public int retrieveIconEnabled(final int... values) {
```
LordMonoxide, line 352: Can you explain what this method does?

Ink230: **how i think icons work**

this.iconFlags_10 has 2 different states. One for human battle menu and one for dragoon battle menu. These are swapped out depending on the turn (and I think controlled by scripts).

So we can't just directly check indexes. We are forced to iterate through and see 'whats there'.  Of course, we can expand the logic and do more overhead checks for what 'state' we are in as a player...but to me that is more lines and if blocks that aren't needed and make the code more dependent on the state implementation.

**the special case**
Item magic and Dragoon magic are never used at the same time. They are currently mutually exclusive. Thus, the same hot key is used for both.

So at a surface level, this form of a method (retrieveIconEnabled) permits one to bind a hot key to multiple actions (of which you may only do one per the current battle menu / icon state). The other method (isIconEnabled) exists to simply check if the icon is enabled, regardless of any state or context.

This is why I used 'retrieve' versus just isEntityEnabled.

**how the method works**
So we iterate from index 0 to the length of the array. For each value in the array, we are then iterating through the values array as well. We check each value to each iconFlag, trying to find any match. Since we know only one state maps to one action, we are only looking for the first match (since there can only be one match).

We immediately exit on the first match or we return the special icon flag of 0 (enums btw (yes, ik its scripted too)).

**extensibility**
Since we do not know what will be refactored with battle menus and/or icons or how exactly mods will interact, I tried to keep it responsive to whatever we decide. But also its simple enough to refactor later if needed. But the main thing is, this implementation doesn't tie us down to whatever we decide to refactor around battle menus or icons.

Since mods may one day be adding new icons to this, we need to not do any fixed/hardcoded comparisons and just be responsive to how the iconFlags and values arrays are currently used.

There is definitely a larger refactor to a) remove battle menu from scripts, b) use enums, c) use a hashmap to not iterate such a nested loop...etc. But atm, the icons are small in size (until Zy adds a 100-icon menu for us). 

**main problem**
If we say 'item magic can be opened in dragoon form' then we have some problems. However, I've paid attention in class.

- We stated we are not making any significant game play changes in Severed Chains
- - Anything that is, should be a mod
- Thus, we should be able to deliver this functionality (and vanilla combat hot keys) as a default vanilla experience
- - Mods will need to override this behaviour
- Luckily...the implementation is very overridable by mod events if we ever get to that point

tl:dr -> I'm meeting the expectation of 'vanilla lod experience' by supporting item menu and dragoon magic menu w/ (the expected) singular hot key for both

PS.
- Changing Additions in-battle => very _QoL_ 'vanilla experience'  :]

LordMonoxide (approved):

PR #1511, opened 2024-07-07
fix: xp gain issue

tredmon:
Partial revert of Pull Request https://github.com/Legend-of-Dragoon-Modding/Severed-Chains/pull/1509/files

### DIFF
--- a/src/main/java/legend/game/inventory/screens/PostBattleScreen.java
+++ b/src/main/java/legend/game/inventory/screens/PostBattleScreen.java
@@ -155,7 +155,7 @@ public class PostBattleScreen extends MenuScreen {
 
           //LAB_8010d9d4
           //LAB_8010d9f8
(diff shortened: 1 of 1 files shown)


LordMonoxide:
We're looking into this at the moment, there's definitely a problem but I don't think this is the right solution, and it isn't specifically related to Meru. Leaving this open for now until we're sure what's going on

Zychronix:
The XP gain is still bugged in one more scenario. The secondayCharSlot loop where it is setting the pendingXp needs to be increased to 9 as well for the fix to be complete.

tredmon:
> The XP gain is still bugged in one more scenario. The secondayCharSlot loop where it is setting the pendingXp needs to be increased to 9 as well for the fix to be complete.

Done

LordMonoxide (approved):

PR #1558, opened 2024-08-11
Unlock Party fixes (Disc 1 & 2), Up script state count to 108, update meta

avionanx:
Fruegel 2 and Gehrich have not been fixed due to texture issues


LordMonoxide (changes requested):
Add comments to the patches to explain what they're doing, everything else looks fine

LordMonoxide (approved):

PR #1568, opened 2024-08-12
Controls Menu | Disable Mouse Input

Ink230:
Play Severed Chains with mouse input disabled. 

### Description
- Hides mouse cursor
- Disables mouse clicks
- Controls do not react to mouse movement, scrolling, or mouse clicks
- Only works if a controller is connected and in use
  - Cursor becomes functional if controller is disconnected

### Motivation
- 💠 Cinematic immersion 💠 
- Controller-only mode (TV setups)
- Avoids niche issues with touchscreen inputs on laptops / TVs / monitors
- In fullscreen border / maximized view the cursor can be hidden
  - Great for single monitor users
  - Helpful for streaming software compatibility
- Click-through overlay programs will work properly
  - Accessibility programs using click-through overlays
- Various mouse based input emulations will work more easily
  - ex: emulating a steering wheel on your mouse as your daily driver
- In the future, modders / us can add specific functionality to controllers that would otherwise break with mouse input
  - ex: controller-designed-only menus / UI
  - ex: rotational-fixed movement UI / game play / cameras (skill issue)

### MenuScreen, ControlHost

Since mouseClick in MenuScreen has a special case for handling mouse clicks w/ regards to hover and focus, the check for mouse input had to be added here, on top of it being added to the super call (since the super method can be used by other screens not going through MenuScreen).

### Window Object
This does not disable window-level mouse inputs.

The mouse is also still tracked by the Window object for its position and is listening for mouse inputs (but they go nowhere currently).

### MenuStack
Atm, could elevate this to MenuStack instead of ControlHost for the current functionality present for menus in-game. But since we may have OtherStack and AdditionalStack that use ControlHost, ControlHost seemed the best to stick the checks.

ex: battle menus, non-menu UIs, modders using ControlHost but not a MenuStack implementation

### getController in Input.java
Is this fine to return the activeController object? It is static and is already edited by many other classes?

### Debugger
Mouse is never hidden in the debugger.

### Renamed GLFW Inputs

image

Very interesting naming scheme, though I think this is standard for games.

Since the free cam one was using disabled but the method was called hideCursor, but I wanted the hidden functionality...renamed the methods + they match the documentation now.

### DIFF
--- a/src/main/java/legend/core/RenderEngine.java
+++ b/src/main/java/legend/core/RenderEngine.java
@@ -32,6 +32,7 @@ import legend.game.combat.Battle;
 import legend.game.input.InputAction;
 import legend.game.modding.coremod.CoreMod;
 import legend.game.types.Translucency;
+import legend.game.input.Input;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.joml.Matrix4f;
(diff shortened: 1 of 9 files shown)


LordMonoxide:
This looks fine but I'm going to defer merging it until we revert the controller code update

Ink230:
Confirmed functionality after rebasing to latest main

LordMonoxide (approved):

PR #1632, opened 2024-09-25
git: Ignore jdk and isos folders

german77:
I don't like to cherry-pick files while creating a commit. Ideally you want to ignore any file that is the result from compilation or setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,11 +6,6 @@
 
 
 
-/.idea/gradle.xml
-/.idea/misc.xml
-
-
-
 # Compiled class file
 *.class
 
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ gradle-app.setting
 
 # Mac
 .DS_STORE
+
+# JDK files
+/jdk*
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -57,3 +52,6 @@ gradle-app.setting
 
 # Mac
 .DS_STORE
+
+# JDK files
+/jdk*
```
Ink230, line 57: Why is this needed?

isos/.gitignore
```diff
@@ -1 +1,2 @@
-*.iso
+*
```
Ink230, line 1: safer, matches what we do in /files

.idea/.gitignore
```diff
@@ -1,3 +1,5 @@
 # Default ignored files
+/gradle.xml
```
Ink230, line 2: logical move

.gitignore
LordMonoxide: Unless you've launched the compiled game from the source directory, this shouldn't be present

german77: I didn't keep the jdk in my PATH.  Since the script downloaded the jdk for me, seemed like a logical move to use the same one that the scripts downloads. 

This isn't strictly necessary so I can remove it.

LordMonoxide: Nah this is fine, I was just curious

LordMonoxide (approved):

Issue #1535, opened 2024-07-27
Barrel Roll during Enemies using "Charging Spirit"


dragoonsouls:
Here i add a Video showing what happened.
Confirmed in this "Death" Enemies.

https://github.com/user-attachments/assets/2382704b-5152-4461-a441-9c546248825e

PR #1639, opened 2024-09-26
Added a fleq condition to calculateStep for camera

### LINKED ISSUE #1285: camera issue in fruegel 1 slam attack
it goes underground

### LINKED ISSUE #1638: Camera spins when Living Statue powers up

theflyingzamboni:
Due to float math rounding discrepancies, sometimes the start and dest values for camera movement take the wrong branch by being unequal when they should be equal. Restoring the code removed in 0b756337 and adding a negative fleq condition to the `val1 > val2` comparison catches these scenarios with an epsilon small enough that it shouldn't catch real differences.

Tested Fruegel slam, Burn Out, and Living Statue Power Up, all worked as expected.

Closes #1638, re closes #1285, #1526

### DIFF
--- a/src/main/java/legend/game/combat/environment/BattleCamera.java
+++ b/src/main/java/legend/game/combat/environment/BattleCamera.java
@@ -2503,7 +2503,13 @@ public class BattleCamera {
 
     //LAB_800dcf48
     if(stepType == 1) {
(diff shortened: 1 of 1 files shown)


avionanx:
Does this close #1535 aswell

LordMonoxide (approved):

PR #1730, opened 2024-10-09
Fixed Queen Fury wake again again probably for real this time

theflyingzamboni:
Wake appeared stair-stepped because the integer angle values lacked the precision necessary for smooth curves at 60fps. Converted remaining math to floats.

Also removed depthOffset render option and moved z offset to transforms. Still unclear why we thought we couldn't do this originally.


LordMonoxide:
See my messages here https://discord.com/channels/307164262063669248/318595603636551701/1293548778325807155

LordMonoxide (approved):

LordMonoxide (approved):

PR #1786, opened 2024-11-04
Fix Dragoon Spells Descriptions

Ink230:
### Damage %'s
- Fixed the dragoon spell descriptions to match the data loaded from file
  - Compared to community spreadsheets and historical forum research
  - Verified the community matched with the actual game data (**they didn't**)
-  image
- Baseline | BaseDamage
  - Obviously you can change the baseline to be whatever, I chose 25% as it matches up with 85% of the existing DMG%s available
  - The 'base damage' calculated mathematically makes sense to be really anything, 25% | 100%...etc
    - This is what is calculated before applyMagicDamageModifier (uses damageMultiplier property) and uses the .multiplier property 
  - My head cannon is that its 25% as that makes the magic damage multiplier reference list make a moderate amount of sense
- Divine Dragoon Spells OP <-> Kongol got nerf'd

### Descriptions
- Motivation
  - I hate inconsistencies
  - Inconsistencies that are inaccurate are worse
  - Inconsistencies that leave out information is the worst
- Terminology
  - Rev -> Revive, % Revive means how much HP you get for reviving
  - Rec -> Recovery, % Recovery means how much HP you get for recovering
  - Cure -> Heals all status ailments
  - 50% Rev, 100% Rec does not mean you get 100% HP when you get revived, it means you end with 50% HP
  - minor -> minor enemies, this isn't what we think of it as ex: Triceratops and Golems are immune to Demon's Gate even though we'd consider them minor enemies...warrants more investigation
  - Leech -> Rose's Astral Drain does 33% leech and spreads it out as a Rec...there's not enough room to describe this but `33% Leech` > `& HP`
- Status
  - Added Stun to Thunder Kids desc to match Rose's Death Dimension description 

### Dev Testing
- Yes, I ensured they all fit
- Rose's Astral Drain has ?2px? to spare on each side but it doesn't look bad


theflyingzamboni (changes requested):
I did this as well in my Disc 1 script overhaul, but I used the actual multipliers as percentages (100%, 300%, 400%, etc.). They convey the same mathematical relationship, but I would personally prefer to go the game multiplier route. It's more accurate to what is in the engine itself, and I think it looks cleaner without decimal points.

These are the changes that I made if you want to glance at it:
https://drive.google.com/drive/folders/11MlYj007iTAvG-ANBAllGmOQo1HJOe8Z?usp=sharing

Regardless, Astral Drain is inaccurate and needs to be changed. I would probably leave it un-detailed on the amount recovered. I don't think there's space available to describe what it actually does, and I'd prefer vague to inaccurate info. If you can figure out how to cram it in, more power to you.

src/main/java/legend/game/Scus94491BpeSegment_8005.java
```diff
@@ -65,20 +65,20 @@ private Scus94491BpeSegment_8005() { }
...
-    "Wind STR 25% All", "Wind STR 75% Single", "Damage Resist 50% Dur 3", "Wind STR 75% All", "STR 50% All",
-    "Light STR 25% All", "Ally Single 100% Rev & Rec", "Ally All 100% Recover", "Light STR 100% All HP", "Wind STR 25% All",
-    "Dark STR 25% Single & HP", "Dark STR 25% All & Fear", "Wind STR 100% Single", "Lethal attack for all", "Dark STR 100% Single",
-    "Thunder STR 50% Single", "Thunder STR 65% Single", "Thunder STR 75% Single", "Thunder STR 100% Single", "Water STR 50% Single",
-    "HP Recv & Cure - All", "Damage Resist 50% Dur 3", "Water STR 50% All", "Water STR 100% Single", "Earth STR 25% All",
+    "Fire STR 50% Single", "Fire STR 25% All", "Fire STR 75% Single", "Fire STR 75% All", "STR 150% Single",
+    "Wind STR 25% All", "Wind STR 75% Single", "Damage Resist 50% Dur 3", "Wind STR 75% All", "STR 100% All",
+    "Light STR 25% All", "Ally 50% Rev, 100% Rec, Cure", "Ally All 50% Rev & Rec, Cure", "Light STR 75% & Allies 100% Rec", "Wind STR 25% All",
+    "Dark STR 50% Single & 33% Leech", "Dark STR 25% All & Fear", "Wind STR 75% Single", "Lethal attack for all - minor", "Dark STR 100% Single",
```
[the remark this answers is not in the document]
Ink230: Reverted to & HP

My brain couldn't come up with any nice ways to cram that info into like 8 characters :D

Ink230:
**Basis**
Think community should vote on switching from 25% basis to 100% basis. Atm, I think its a non-blocker as is to get more dmg% numbers in SC for now. I have no preference cause I can do math but will note most of the fandom is using 25% basis...whether they realize it or not.

**Demon's Gate**
I love the description in your doc: `Drags minor enemies to Hell`
If Monoxide approves and Drew is on board...I'm on board.

I wanted to find something to capture 'light', 'small' over just 'minor' but no idea.

**Programmatic Generation**
Monoxide brought it up to make these descriptions generated programmatically.
I think the existing dragoon spells logic makes this difficult as the flags don't directly describe the full behaviour, especially for the utility spells.

I think there's a) dragoon spells rework internally to put all of the logic in Java w/ proper flags/columns, b) an adaptive UI or font to properly describe the descriptions, and c) community needs to agree on a format.

theflyingzamboni (dismissed):
I'm fine with the number format being a non-blocking decision.

src/main/java/legend/game/Scus94491BpeSegment_8005.java
```diff
@@ -65,20 +65,20 @@ private Scus94491BpeSegment_8005() { }
...
 
   public static final String[] spellCombatDescriptions_80052018 = {
-    "Fire STR 50% Single", "Fire STR 25% All", "Fire STR 75% Single", "Fire STR 175% All", "STR 100% Single",
-    "Wind STR 25% All", "Wind STR 75% Single", "Damage Resist 50% Dur 3", "Wind STR 75% All", "STR 50% All",
-    "Light STR 25% All", "Ally Single 100% Rev & Rec", "Ally All 100% Recover", "Light STR 100% All HP", "Wind STR 25% All",
-    "Dark STR 25% Single & HP", "Dark STR 25% All & Fear", "Wind STR 100% Single", "Lethal attack for all", "Dark STR 100% Single",
-    "Thunder STR 50% Single", "Thunder STR 65% Single", "Thunder STR 75% Single", "Thunder STR 100% Single", "Water STR 50% Single",
-    "HP Recv & Cure - All", "Damage Resist 50% Dur 3", "Water STR 50% All", "Water STR 100% Single", "Earth STR 25% All",
+    "Fire STR 50% Single", "Fire STR 25% All", "Fire STR 75% Single", "Fire STR 75% All", "STR 150% Single",
+    "Wind STR 25% All", "Wind STR 75% Single", "Damage Resist 50% Dur 3", "Wind STR 75% All", "STR 100% All",
```
LordMonoxide, line 69: Is there room to put "3 Turns" instead of "Dur 3"?

LordMonoxide:
@Ink230 25% is good because it minimizes changes. I'm fine with tfz's description for demon's gate.

src/main/java/legend/game/Scus94491BpeSegment_8005.java
Ink230: Yes

Ink230:
Done both

PR #1815, opened 2024-11-24
Fixed sparse rain effect

### LINKED ISSUE #1807: Melbu Gen 3 rain gets too sparse partway in

theflyingzamboni:
Java modulo strikes again.

Also added adjustment for raindrop count based on screen width. Would like other opinions on if it seems like a similar density, or if the counts were already overestimated and just rendering off-screen before merging.

Fixes #1807


LordMonoxide:
I thought about this for a while and I think the aspect ratio thing is fine, but the main render batch is private for a reason, we don't want mods being able to arbitrarily change the main projection

LordMonoxide (approved):

PR #1840, opened 2024-12-09
Fixed Ch 2 Giganto Home unlocked party changes (closes #1772)

### LINKED ISSUE #1772: Giganto Home - Party member cloning before Gehrich boss fight
Enter fight room with Haschel in slot 0.
Slot 2 member is moved to inactive.
Slot 1 member is moved to slot 2.
A second instance of Haschel is created, occupying both slot 0 and slot 1.
Both Haschels remain in the party even after Kongol joins.

Other observations on multi-instanced characters (probably known/expected behavior):
Duplicated characters share addition, normal, and dragoon level EXP.
When both/all active, they each generate addition and dragoon level EXP, but will only get one share of normal EXP, wasting 1-2 active slots' worth of normal EXP.
Active addition is also shared, so if you change it mid-battle for one, it changes for the other(s).

warblgarbl:
#### **Retail**
When replacing Haschel, slot 0 is ignored. It will attempt to replace him in slot 1 or 2 with Meru or Shana, depending on the other slot's current occupant.
#### **Unlock party fix**
Replace Haschel in any slot with the first available character on standby. Only change Haschel's party flag.

#### **Unlock party fix**
Only change Haschel's party flag.

#### **Unlock party fix**
Force Dart into active party (slot 0) when everyone else is supposed to leave the party.

closes #1772


LordMonoxide (approved):
Looks good, if you're satisfied with this, fix the conflict and I'll merge

warblgarbl:
All set

LordMonoxide (approved):

PR #1859, opened 2024-12-18
Updated PolyBuilder so vertices can inherit color as well

theflyingzamboni:
Someone please double-check that I didn't delete any of the wrong calls.


src/main/java/legend/core/opengl/PolyBuilder.java
```diff
@@ -199,7 +205,7 @@ public MeshObj build() {
   private static class Vertex {
     private final Vector3f pos = new Vector3f();
     private final Vector2f uv = new Vector2f();
-    private final Vector3f colour = new Vector3f();
+    private Vector3f colour = new Vector3f();
```
LordMonoxide, line 208: Is there a reason these need to be non-final? It looks like we're always instantiating them in addVertex anyway, they should just be instantiated here and set in addVertex

src/main/java/legend/game/combat/effects/GradientRaysEffect24.java
```diff
@@ -183,28 +183,22 @@ private void renderGradientRay(final EffectManagerData6c<EffectManagerParams.Voi
       if(effect.type_1c == 1) {
         builder
           .addVertex(xy1.x, xy1.y, 0)
-          .monochrome(0)
           .addVertex(xy2.x, xy2.y, 0)
           .rgb(r, g, g)
```
LordMonoxide, line 187: Uhh, is this right?

src/main/java/legend/core/opengl/PolyBuilder.java
theflyingzamboni: Don't think they need to be non-final. Vertex clut field was non-final when I originally made the clut/vram per-vertex change, so I just copied it I guess lol

src/main/java/legend/game/combat/effects/GradientRaysEffect24.java
theflyingzamboni: Yes

image

LordMonoxide (approved):

PR #1872, opened 2024-12-21
Improve Encounters

Ink230:
### Description
- Debugger BattleStage selector did not work for Contact Encounters
- Debugger BattleStage logic was in multiple places
- Encounter flows were divergent and had a mix of duplicated logic in multiple places

### tl:dr
- Debugger BattleStageId control works for all encounter types with all stages
- Encounter flow logic consolidated, streamlined, and updated
- Removed 0 being false and an int
- Encounter related mod events will now work for all encounter types

---

### 0a24d0f

**Good**
- Unnest if else logic
- Conform the nested else if to the rest of the method's early return convention
- Make ternary the disjointed logic checks for encounterId_800bb0f8
  - Remove unnecessay encounterId blank finalization variable
- 0x1ff -> stageData_80109a98.length as this is what its checking for and what its counterpart does in handleEncounters flow

**Bad**
- Don't like that the previous state of encounterId_800bb0f8 is used as a fallback / error as it adds error surface area
- StageData is relying on global state and submap controller script sync to be correct (non-ideal)
- Still a secondary but unjustifiable reason to keep this flow separate from parts of handleEncounter flow
- Index IDs are non-ideal

### 943e407

**Good**
- battlestage debugger control works for all encounter types
- battlestage debugger check is checked only once now instead of 3 times in mutliple places
- battlestage debugger startEncounter is now 1:1 with scriptMapTransition
  - Compared to these implementations being the same intent but with diverged implementations
  - startEncounter's submap battle checks *should* be movable to somewhere else pending script adjustments with scriptMapTransition
- battlestage mod events now works in one place without any other wonk in multiple places
- Easier future rewrite for whatever Script Translation Layer | BattleMediator | Battle State Machine rewrite occurs

**Bad**
- prepareEncounter
  - Ugly if checks not resolved by proper contesting and flows
    - But this is where they belong
  - Can't actually start encounter id of 0 via debugger
  - Can't mod event battle id of 0
- scriptMapTransition
  - Same as above, ugly if check for combat condition
    - Can't resolve w/o some rewrite in whatever Script Translation Layer | BattleMediator | Battle State Machine rewrite occurs
    - scriptMapTransitionBattle should exist and be used in scripts when we want to do battle map transitions
    - But scriptMapTransition is used for it all from scripts

### 0e7bc01

**Good**
- Recover all prior existing functionality in prepareEncounter that was missed out in the previous commit
- Much cleaner
- Debugger, Mod Events, Contact Encounters, and Walkable Encounters should now all work without issue

**Bad**
- Not sure what precedence to give Config vs Mod Event...I presume debugger should take ultimate priority

Sent from Android

### DIFF
--- a/src/main/java/legend/game/submap/RetailSubmap.java
+++ b/src/main/java/legend/game/submap/RetailSubmap.java
@@ -409,18 +409,15 @@ public class RetailSubmap extends Submap {
   }
 
   @Override
-  public void prepareEncounter(final int targetEncounterId) {
+  public void prepareEncounter(final Integer targetEncounterId) {
     final var sceneId = encounterData_800f64c4[this.cut].scene_00;
     final var scene = sceneEncounterIds_800f74c4[sceneId];
-    final var encounterId = targetEncounterId == 0 ? scene[this.randomEncounterIndex()] : targetEncounterId;
-    final var battleStageId = encounterData_800f64c4[this.cut].stage_03;
+    final var encounterId = targetEncounterId == null ? scene[this.randomEncounterIndex()] : targetEncounterId;
+    final var battleStageId = encounterData_800f64c4[this.cut].stage_03 == 0 && battleStage_800bb0f4 > -1 ? battleStage_800bb0f4 : encounterData_800f64c4[this.cut].stage_03;
 
     final var generateEncounterEvent = EVENTS.postEvent(new SubmapGenerateEncounterEvent(encounterId, battleStageId, this.cut, sceneId, scene));
     encounterId_800bb0f8 = generateEncounterEvent.encounterId;
-
-    if(generateEncounterEvent.battleStageId != 0) {
-      battleStage_800bb0f4 = generateEncounterEvent.battleStageId;
-    }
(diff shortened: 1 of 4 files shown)


src/main/java/legend/game/submap/RetailSubmap.java
```diff
@@ -408,15 +409,19 @@ public int getEncounterRate() {
   }
 
   @Override
-  public void generateEncounter() {
+  public void prepareEncounter(final Integer targetEncounterId) {
```
LordMonoxide, line 412: I don't like using the integer object here so that it can be null or an integer. Thinking from a modder's perspective, they'll have to know that it can be null and I can see this easily leading to null pointer exceptions. What if we have a second `prepareEncounter` overload that takes no param and then calls `prepareEncounter(whateverValue)`?

Ink230: Overload used with updated correct logic and flow

LordMonoxide (approved):

PR #1883, opened 2024-12-23
Shop Registry

Ink230:
### Questions

- **Is this done right?**
  - LodItem and LodEquipment do different things compared to each other
  - Translation helper methods not present (not sure if needed?)
  - Not sure what from the other registries is needed for a Shop
    - ShopEntry Interface?
    - Shops are a data structure in Java, not a file
      - Up to the mod to source the Shops data?
      - Which is why I put it in a LinkedHashMap in lodmod under Shop?
        - Where else and how should this live otherwise? 
   - ShopStruct40 was used as the backing type, implements RegistryEntry
     - Didn't know if I should move this to 'Inventory'
     - But all the other Shop stuff is in 'Inventory'
     - Or if we create a new type and abandon ShopStruct40? Idk. 
- **As far as I can get without touching scripts**
  - SMap dependency on index still exists
  - ShopScreen.java can't be rewritten yet
  - Inner crisis that shopType exists as a property and we don't just handle Inventory or Items gracefully together

I assume the submap script is going to callback to Java-land and *logic happens* to know what shop to open for the current submap / context / reasons but atm the submap script(s) dictate by an index which shop to open?

I'm down to get dirty in the script mines to finish this off but will need lessons™️

ps: shop ids were matched based on community data and were tested for a few disc 3 and disc 1 shops


src/main/java/legend/game/submap/SMap.java
```diff
@@ -300,70 +301,70 @@ public class SMap extends EngineState {
...
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
-    new ShopStruct40(0),
+    LodShops.BALE_EQUIPMENT_CASTLE.get(),
```
LordMonoxide, line 304: Resolving these delegates and storing references to the shop instances will cause everything to break when the registries are dumped and reloaded. They'll point to the shops that were loaded the first time that SMap was loaded rather than what's currently loaded.

src/main/java/legend/game/types/ShopStruct40.java
```diff
@@ -1,10 +1,11 @@
...
 
 import legend.game.inventory.InventoryEntry;
+import org.legendofdragoon.modloader.registries.RegistryEntry;
 
 import java.util.function.Supplier;
 
-public class ShopStruct40 {
+public class ShopStruct40 extends RegistryEntry {
   public final int shopType_00;
   public final Supplier<InventoryEntry>[] items_00;
```
LordMonoxide, line 10: We'll need to lock down this array, right now a mod could overwrite the suppliers
i.e. `shop.items_00[0] = whatever;`

src/main/java/legend/lodmod/LodShops.java
```diff
@@ -0,0 +1,50 @@
+package legend.lodmod;
+
+import legend.core.GameEngine;
+import legend.game.types.ShopStruct40;
+import org.legendofdragoon.modloader.registries.RegistryDelegate;
+
+public class LodShops {
+  private LodShops() { }
+
+  public static final RegistryDelegate<ShopStruct40> BALE_EQUIPMENT_CASTLE = GameEngine.REGISTRIES.shop.getEntry(LodMod.id("bale_equipment_castle"));
```
LordMonoxide, line 10: Instead of having a separate Shops class, the shops should be instantiated here and registered using a registrar. See LodItems for how it's done.

Ink230:
- Made the implementation from LodItems instead of LodEquipment
- Made ShopStruct40 property items_00 private and accessible by accessors only
- SMap shops variable now uses RegistryDelegate[]
- X Shops class, ✔ our data for shops in LodShops
  - row-order normalization still unavoidable as of yet

LordMonoxide:
The code looks good now. Two questions, do we still need the old shops array, and do we need the extra empty shops that were padding out the array?

Ink230:
**Still need shops array**

No
- ShopScreens, SMap...etc, will be fully swappable by LodMod-like mods
  - Meaning any hardcoded dependencies we put in existing classes for LodMod won't matter in the future
  - Ex: Usage of LodMod in the latest commit for the REGISTRIES.shop to access the shop by SHOP_IDS index
- We remove index-dependencies in the script and use string IDs
  - Aka, the scripts become a source of truth by demanding a string ID to be present?
  - Ex: ShopScreens -> instead of a shopId from the scripts, it should pass a string ID to load
  - Then we don't need to use a) indices, b) LodMod properties and can only query the REGISTRY, which is mod-agnostic
- We're ok with any of the above being a temporary solution until RB4-proper

**Empty Shop Padding**
- Doesn't hurt anything right now
- No way to confirm other than through scripts or running through the all the shops to confirm which ones are used where
- Potentially this needs a system to make sure
  - The index passed in from scripts is in range of the length of the shop registry? This is irrelevant once we lose reliance on indices
  - The id is present (however, crashing and not handling this is probably best for modding to avoid silent invalid states?)

LordMonoxide (approved):

PR #1919, opened 2024-12-30
Adjust camera for Haschel solo punch cutscene

### LINKED ISSUE #1918: Submap scissoring issue - Haschel close-up cutscene breaking stone barrier
The cutscene takes place to the left of the submap, so nothing renders due submap scissoring.
The submap is visible to the right in ultrawide resolutions.

warblgarbl:
Moves the camera above the submap rather than to its left. This places it within the scissor's boundaries to render the scene as expected (still cropped to the submap's width).

Closes #1918


warblgarbl:
```
  private FlowControl scriptSetCameraOffset(final RunningScript<?> script) {
    final int x = script.params_20[0].get();
    final int y = script.params_20[1].get();
    final int mode = this.scriptSetOffsetMode_800f7e50;

    if(mode == 1 || mode == 2) {
      //LAB_800e695c
      this.setGeomOffsetIfNotSet(3 - vsyncMode_8007a3b8, x, y);
    }

    if(mode != 1) {
      this.setScreenOffsetIfNotSet(3 - vsyncMode_8007a3b8, x, y);
    }

    //LAB_800e6988
    this.scriptSetOffsetMode_800f7e50 = 0;
    return FlowControl.CONTINUE;
  }
  ```
Mode 0 is used, so the only thing that moves relative to the camera is the background.
Patch changes the direction from horizontal to vertical.

LordMonoxide (approved):

PR #1927, opened 2025-01-01
Fix for breath waves not looping correctly and various audio enhancements

Illeprih:


LordMonoxide (changes requested):
Code changes look good, only thing that needs to change is adding translation support for the dropdowns of all 3 configs. Take a look at `EnumConfigEntry` (you can probably just extend that instead of implementing the dropdowns yourself)

LordMonoxide (approved):

PR #1937, opened 2025-01-04
Implement screen capture effects in opengl and fixed related bugs

### LINKED ISSUE #1563: Screen captures not rendering

theflyingzamboni:
-Fixed bug from transforms not being loaded into GTE when effects rendered.

Closes #1563

### DIFF
--- a/patches/scripts/DRGN0/5466/1.diff
+++ b/patches/scripts/DRGN0/5466/1.diff
@@ -15,7 +15,21 @@
  mov 0x14, stor[9]
  wait stor[9]
  LABEL_22:
-@@ -1974,7 +1978,7 @@
+@@ -1764,10 +1768,12 @@
+ call 551, stor[21], 0xffffffff, 0x0, 0x0, 0x0
+ call 627, stor[21], 0xffffff38
+ call 589, stor[21], 0x0
+-call 897, stor[22], 0x14400, 0x0, 0x0, 0x80, 0x60, 0x1
+ call 42, 0x1, 0x0, 0x0, 0x0, stor[11]
+ call 42, 0x1, 0x0, 0x1, 0x0, stor[12]
+ call 42, 0x1, 0x0, 0x2, 0x0, stor[13]
++mov 0x1, stor[9]
++wait stor[9]
++call 897, stor[22], 0x14400, 0x0, 0x0, 0x80, 0x60, 0x1
+ call 545, stor[22], 0xffffffff, stor[11], stor[12], stor[13]
+ call 547, stor[22], 0xffffffff, 0x0, 0x800, 0x0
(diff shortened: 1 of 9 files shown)


src/main/java/legend/game/combat/effects/ScreenCaptureEffect1c.java
```diff
@@ -35,7 +45,62 @@ public class ScreenCaptureEffect1c implements Effect<EffectManagerParams.VoidTyp
...
+      builder.size(w, h);
+      builder.internalFormat(GL_RGBA16_SNORM);
+      builder.dataFormat(GL_RGBA);
+      builder.dataType(GL_UNSIGNED_BYTE);
+      builder.magFilter(GL_NEAREST);
+      builder.minFilter(GL_LINEAR);
+    });
+
+    final ByteBuffer data = BufferUtils.createByteBuffer(w * h * 4);
+    RENDERER.getLastFrame().use();
```
LordMonoxide, line 74: Trying to wrap as much opengl stuff as possible, can you move the use and glGetTexImage to a method on Texture called getData? Can also probably add a check that the buffer is the correct size (Texture has width/height properties), iirc the correct property to get buffer size is either limit or capacity

patches/scripts/DRGN0/5466/1.diff
```diff
@@ -15,7 +15,21 @@
...
-@@ -1974,7 +1978,7 @@
+@@ -1764,10 +1768,12 @@
+ call 551, stor[21], 0xffffffff, 0x0, 0x0, 0x0
+ call 627, stor[21], 0xffffff38
+ call 589, stor[21], 0x0
+-call 897, stor[22], 0x14400, 0x0, 0x0, 0x80, 0x60, 0x1
+ call 42, 0x1, 0x0, 0x0, 0x0, stor[11]
+ call 42, 0x1, 0x0, 0x1, 0x0, stor[12]
+ call 42, 0x1, 0x0, 0x2, 0x0, stor[13]
++mov 0x1, stor[9] # Moved camera updates before screen capture allocation and added 1 frame wait to get screen to capture on the correct frame
```
LordMonoxide, line 26: Do comments work when it's not at the start of the line, or is this going to be inserted into the script?

src/main/java/legend/core/opengl/Texture.java
```diff
@@ -185,6 +187,20 @@ public void dataInt(final int x, final int y, final int w, final int h, final in
     }
   }
 
+  public void getData(final ByteBuffer data) {
+    if(data.capacity() == this.width * this.height * 4) {
```
LordMonoxide, line 191: If you invert this if statement you won't need an else branch since the throw will exit

LordMonoxide (approved):

PR #1949, opened 2025-01-05
Unify font renderer methods and add font size option (closes #1944)

### LINKED ISSUE #1944: Add font renderer that supports resizing text
This will be needed for options menu improvements

LordMonoxide:


Zychronix:
Code looks good! Tested the following: Various menus, character status, in-battle text boxes, battle cutscene text, world map city/destination display,

PR #1962, opened 2025-01-05
Slugify campaign names (closes #1933)

### LINKED ISSUE #1933: Using : in path or leaving extra spaces at the end breaks campaign creation
If you use characters that aren't allowed in paths, the game will crash when you start a new game. Campaigns should do what saves do and slugify the name

LordMonoxide:
This will replace any character that's unsafe for paths with a `_`. The actual name is now saved in the campaign config so what the user typed into the name box is still displayed as normal. This matches the behaviour of save names.

Leading and trailing whitespace is now stripped in campaign/save names as well.


Ink230:
- created new saves with unsafe characters
- loaded said saves
- loaded previous campaign saves
- started new campaign
- saved saves
- loaded said saved saves

nothing corrupted or seems weird

PR #1966, opened 2025-01-05
discussion listmenu mp fix

Ink230:
1. The changediff is what clears the stale data and is just there to illustrate the problem
2. Stale data is being populated from enemy turns (aka, garbage data not so much stale data) and is not consistently cleaned up
3. Case 2 needs to be refactored and split each check into a doCase2 for AdditionListMenu, SpellListMenu, and ItemListMenu
4. Likely even the whole tick method and all other cases

- Do you want me to hunt down all leafs in the flow to find where spell_94 and item_d4 should be set to null?
- Refactor just the resetting of these two properties to null before/after/during this Case 2 / onSelection in ListMenu proper?
- Use this fix for now / put in onSelection and reset spell_94 in itemlist and spelllist
- Implement an MP check for ItemListMenu and SpellListMenu (where only SpellListMenu will have any functionality)?

Technically item_d4 doesn't need to be reset to null but it is used in the same manner as spell_94 and should follow the same conventions. item_d4 is also suffering from the same stale/garbage data but since there's no validations (like MP) its not a concern (yet).

### DIFF
--- a/src/main/java/legend/game/combat/ui/ListMenu.java
+++ b/src/main/java/legend/game/combat/ui/ListMenu.java
@@ -102,6 +102,10 @@ public abstract class ListMenu {
   protected abstract int handleTargeting();
   public abstract void getTargetingInfo(final RunningScript<?> script);
 
+  protected boolean canUse() {
+    return true;
+  };
+
   private void initObjs() {
     if(this.menuObj == null) {
       final QuadBuilder builder = new QuadBuilder("Spell/item menu");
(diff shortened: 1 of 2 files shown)


Ink230:
changes pushed to branch with canUse
- this sidesteps the state problem and removes the need for a null check
- itemlist and additionlist always return true for canUse but are never called as there's no need for them in there (yet)

fixes the original issue, fixes garbage state in the 2 children interfering with each other

LordMonoxide (approved):

PR #1980, opened 2025-01-19
feat: cutscene skips on double tap

### LINKED ISSUE #1917: Restrict cutscene skips
- Restrict the skip button to a single key
- If a controller button is pressed, show a message that says "Press <button> to skip"
- If a keyboard key is pressed, show a message that says "Press <key> to skip"

Controller button should NOT be A/cross, since players may already be pressing that. Maybe Y/triangle. Keyboard key can be enter.

aharonphytech:
Resolves #1917

### DIFF
--- a/src/main/java/legend/game/fmv/Fmv.java
+++ b/src/main/java/legend/game/fmv/Fmv.java
@@ -16,6 +16,7 @@ import legend.core.opengl.fonts.FontManager;
 import legend.core.opengl.fonts.TextStream;
 import legend.core.spu.XaAdpcm;
 import legend.game.EngineStateEnum;
+import legend.game.i18n.I18n;
 import legend.game.input.Input;
 import legend.game.input.InputAction;
 import legend.game.types.Translucency;
@@ -278,6 +279,9 @@ public final class Fmv {
 
   private static void setSkipText(final String text, final InputSource inputSource) {
     synchronized(skipTextLock) {
+      if(skipText != null) {
+        skipText.delete();
+      }
       skipText = font.text(stream -> stream.text(text));
     }
     currentInputSource = inputSource;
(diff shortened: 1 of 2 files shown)


src/main/java/legend/game/fmv/Fmv.java
```diff
@@ -244,7 +266,53 @@ public static void playCurrentFmv(final int fmvIndex, final EngineStateEnum afte
...
+      Translucency.HALF_B_PLUS_HALF_F.setGlState();
+      skipText.setColour(1.0f, 1.0f, 1.0f);
+      skipText.draw(10, 10, 0.5f, 0.5f);
+      glDisable(GL_BLEND);
+    }
+  }
+
+  private static void setSkipText(final String text, final InputSource inputSource) {
+    synchronized(skipTextLock) {
+      skipText = font.text(stream -> stream.text(text));
```
LordMonoxide, line 281: This is allocating memory on the GPU and it needs to be freed any time the `skipText` pointer is overwritten or nulled.

`skipText.delete()`

src/main/java/legend/game/fmv/Fmv.java
```diff
@@ -244,7 +266,53 @@ public static void playCurrentFmv(final int fmvIndex, final EngineStateEnum afte
...
+    }
+    currentInputSource = inputSource;
+    skipTextFramesRemained = 60;
+  }
+
+  private static void handleSkipText() {
+    if(isKeyboardInput) {
+      setSkipText("Press ENTER to skip", InputSource.KEYBOARD);
```
LordMonoxide, line 289: These 3 new strings should go through the translation system. You can define the strings in `resources/lod_core/lang/en.lang` and use `I18n.translate`.

e.g.
```
setSkipText(I18n.translate("lod_core.fmv.skip_keyboard"), InputSource.KEYBOARD);
```

LordMonoxide (changes requested):

src/main/java/legend/game/fmv/Fmv.java
aharonphytech: Done

src/main/java/legend/game/fmv/Fmv.java
aharonphytech: Done

LordMonoxide (approved):

PR #2010, opened 2025-02-02
Render overhaul part 87

LordMonoxide:
### Change A
(B+F)/2 translucency for TMDs was just broken, used 100% alpha instead of 50%

### Change B
Added depth culling comparison option to queued models. I had changed it from GL_LESS to GL_LEQUAL a while ago to fix UI rendering, and that actually fixed some other things (like fixing Haschel's face textures not rendering) but also broke other things (chest shadow rendering overtop of open lid). I'm still defauling to GL_LEQUAL, but using GL_LESS for sobj models. No idea if this is breaking/fixing anything else. Turns out retail models are a mess of overlapping geometry.

### Change C
Overhauled render engine to give it full control over rendering individual layers of models. Before, it worked like this:
1. Render all opaque layers (opaque)
2. Render all translucent layers with translucent texture bit disabled (opaque)
3. Render all translucent layers with translucent texture bit enabled (translucent)

This was causing Haschel's face texture to get depth culled because it was rendered at the same depth as his head model. The head was always rendered first. The fix was two parts:
1. Change the TMD loader to load meshes in the OPPOSITE order to mimic the order they would be rendered when sent to the GPU's queue (LIFO)
2. Combine steps 1 and 2 of the previous list, so the layers are rendered in order (instead of `all layers that meet X criteria` then `all layers that meet Y criteria`

### Change D
Made freecam work in old renderer for testing

### DIFF
--- a/gfx/shaders/battle_tmd.fsh
+++ b/gfx/shaders/battle_tmd.fsh
@@ -92,7 +92,7 @@ void main() {
   outColour.rgb *= recolour;
 
   // The or condition is to disable translucency if a texture's pixel has alpha disabled
-  if(translucencyMode == 0 && (!textured || outColour.a != 0)) { // (B+F)/2 translucency
+  if(translucencyMode == 1 && (!textured || outColour.a != 0)) { // (B+F)/2 translucency
     outColour.a = 0.5;
   } else {
     outColour.a = 1.0;
--- a/src/main/java/legend/core/QueuedModel.java
+++ b/src/main/java/legend/core/QueuedModel.java
@@ -16,6 +16,7 @@ import java.nio.FloatBuffer;
 import java.util.Arrays;
 
 import static legend.core.GameEngine.GPU;
+import static org.lwjgl.opengl.GL11C.GL_LESS;
 
 public abstract class QueuedModel<Options extends ShaderOptionsBase<Options>, T extends QueuedModel<Options, T>> {
   protected final RenderBatch batch;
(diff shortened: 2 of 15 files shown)


gfx/shaders/battle_tmd.fsh
```diff
@@ -92,7 +92,7 @@ void main() {
   outColour.rgb *= recolour;
 
   // The or condition is to disable translucency if a texture's pixel has alpha disabled
-  if(translucencyMode == 0 && (!textured || outColour.a != 0)) { // (B+F)/2 translucency
+  if(translucencyMode == 1 && (!textured || outColour.a != 0)) { // (B+F)/2 translucency
```
LordMonoxide, line 95: This was just wrong

src/main/java/legend/core/QueuedModel.java
```diff
@@ -41,6 +43,9 @@ public abstract class QueuedModel<Options extends ShaderOptionsBase<Options>, T
   final Texture[] textures = new Texture[32];
   boolean texturesUsed;
 
+  int opaqueDepthComparator;
```
LordMonoxide, line 46: Allows model queues to specify their depth comparison for depth culling

dragoonsouls:
> ### Change B
> Added depth culling comparison option to queued models. I had changed it from GL_LESS to GL_LEQUAL a while ago to fix UI rendering, and that actually fixed some other things (like fixing Haschel's face textures not rendering) but also broke other things (chest shadow rendering overtop of open lid). I'm still defauling to GL_LEQUAL, but using GL_LESS for sobj models. No idea if this is breaking/fixing anything else. Turns out retail models are a mess of overlapping geometry.

Now you get why i did my overlapping geometry algorithm in my tool... hehehehe... if not everything (specially face features are all behind the main head geometry).

PR #2091, opened 2025-02-15
Fix FMV projection aspect ratio

### LINKED ISSUE #2064: FMVs use wrong aspect ratio
Vertical pixel count is wrong, circles look oval-shaped. Retail has more pronounced letterboxing.

Seems to require the inverse of the game-engine's usual width squish (368 -> 320). Dividing the vertical count by 1.15 (368 / 320) turns the ovals back into circles.

(Uploading comparisons later)

warblgarbl:
Closes #2064


LordMonoxide:
Thanks for the contribution but I ended up reworking FMVs to use the render engine rather than raw opengl and fixed this while I was in there

PR #2105, opened 2025-02-18
Status Debugger

avionanx:
The goal for RB3 here is to provide a very quick way of inflicting conditions on yourself or enemies, so this removes:

- Having to stockpile yourself with status condition items
- Waiting for an enemy to cast specific attack
- Having to look up which enemy does what move, etc.

This debugger currently does not provide menu blocking and can't combat
It's suited for use when battle is idle, which is when battle HUD is up. Setting status during an addition or deff is untested.

For RB4, if adding registries to status conditions is on the table, we can replace dropdown with a text field.


src/main/java/legend/game/debugger/CombatDebuggerController.java
```diff
@@ -108,6 +121,12 @@ public ListItem fromString(final String string) {
       this.displayStats(index);
     });
 
+    this.statusCondition.getItems().add("None");
+    for(int i = BattleHud.ailments_800fb3a0.length - 1; i >= 0; i--){
```
LordMonoxide, line 125: Needs space between `){` (applies elsewhere in this file too)

LordMonoxide (approved):

PR #2108, opened 2025-02-18
Polish multiple menus

### LINKED ISSUE #296: Menu Wrapping
If you press down at the bottom, the cursor does not go back to the top and vice versa.

### LINKED ISSUE #1754: Sell items characters doesn't change
When you go to sell items, it's supposed to change to show you what characters can equip said items so you can be sure you're not selling something you actually need.  However, it seems to lock to whoever the first characters are (so if Dart and Lavitz can equip the first items, it'll only show them for all).

### LINKED ISSUE #2112: Mouse wheel still scrolling when mouse disabled = true

warblgarbl:
Add list wrapping
Fix input sounds
Fix compatible character filtering in shop menus

Closes #296
Closes #1754
Closes #2112

### DIFF
--- a/src/main/java/legend/game/inventory/screens/AdditionsScreen.java
+++ b/src/main/java/legend/game/inventory/screens/AdditionsScreen.java
@@ -32,8 +32,6 @@ import static legend.game.Scus94491BpeSegment_800b.characterIndices_800bdbb8;
 import static legend.game.Scus94491BpeSegment_800b.gameState_800babc8;
 import static org.lwjgl.glfw.GLFW.GLFW_KEY_END;
 import static org.lwjgl.glfw.GLFW.GLFW_KEY_HOME;
-import static org.lwjgl.glfw.GLFW.GLFW_KEY_PAGE_DOWN;
-import static org.lwjgl.glfw.GLFW.GLFW_KEY_PAGE_UP;
 import static org.lwjgl.glfw.GLFW.GLFW_MOUSE_BUTTON_LEFT;
 
 public class AdditionsScreen extends MenuScreen {
@@ -49,7 +47,8 @@ public class AdditionsScreen extends MenuScreen {
   private final MenuAdditionInfo[] additions = new MenuAdditionInfo[9];
 
   /** Allows list wrapping, but only on new input */
-  private boolean allowWrap = true;
+  private boolean allowWrapX = true;
+  private boolean allowWrapY = true;
 
   public AdditionsScreen(final Runnable unload) {
     this.unload = unload;
@@ -92,7 +91,7 @@ public class AdditionsScreen extends MenuScreen {
 
           if(this.charSlot > 0) {
             this.scroll(this.charSlot - 1);
-          } else {
+          } else if(characterCount_8011d7c4 > 1 && this.allowWrapX) {
             this.scroll(characterCount_8011d7c4 - 1);
           }
         }
@@ -102,7 +101,7 @@ public class AdditionsScreen extends MenuScreen {
 
           if(this.charSlot < characterCount_8011d7c4 - 1) {
             this.scroll(this.charSlot + 1);
-          } else {
+          } else if(characterCount_8011d7c4 > 1 && this.allowWrapX) {
             this.scroll(0);
           }
         }
@@ -238,7 +237,7 @@ public class AdditionsScreen extends MenuScreen {
     if(this.selectedSlot > 0) {
       playMenuSound(1);
       this.selectedSlot--;
-    } else if(this.allowWrap) {
+    } else if(this.allowWrapY) {
       playMenuSound(1);
       this.selectedSlot = 6;
     }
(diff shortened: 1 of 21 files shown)


LordMonoxide:
Needs wrapping on:
- Options/new campaign/mods list (all VerticalLayoutScreen)
- MessageBoxScreen

Inconsistencies noticed:
- In the "Use Item" menu, using an item when the list is scrolled all the way down leaves a blank space instead of scrolling up (contrast with "Inventory" menu)
- Addition menu allows scrolling to empty slots (I think this is retail behaviour but we may want to change it?)
- Char swap screen ignores repeat input (holding down a direction will only scroll once)
- Char swap screen doesn't have wrapping, not as important as other screens but it'd be nice to wrap up/down on the left list and wrap from first<->last on the right list since it wraps from the first row right -> to the second row

warblgarbl:
> Needs wrapping on:
> 
> * Options/new campaign/mods list (all VerticalLayoutScreen)
> * MessageBoxScreen
* Oops.
* Single press or repeat key?

> Inconsistencies noticed:
> 
> * In the "Use Item" menu, using an item when the list is scrolled all the way down leaves a blank space instead of scrolling up (contrast with "Inventory" menu)
Got it.
> * Addition menu allows scrolling to empty slots (I think this is retail behaviour but we may want to change it?)

Yes it's retail behavior, and I'm gonna say we probably shouldn't change it. I've seen speed runners prep their slot selection as they're cycling through to the intended character.

> * Char swap screen ignores repeat input (holding down a direction will only scroll once)
> * Char swap screen doesn't have wrapping, not as important as other screens but it'd be nice to wrap up/down on the left list and wrap from first<->last on the right list since it wraps from the first row right -> to the second row

No problem, wasn't too sure how desired it was on char swap. Should it also use the wrap halting for either up/down or left/right?

LordMonoxide:
For wrapping of VerticalLayout/MessageBox, keep it consistent and wrap on press, not repeat. Same for char swap, keep it consistent.

For additions, that's fine, we can leave it as vanilla if the speedrunners expect that behaviour.

warblgarbl:
All set, but one thing we should probably verify on the additions part is whether retail allows up/down selection while on an archer page and/or whether that would matter to a speedrunner.

Then if we do allow it, do we mask the menu sound for a "normal" experience?

LordMonoxide:
Crash when using an item that would leave less than 5 items on the "Use Item" screen:
```
java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 4
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at legend.game.SItem.renderMenuItems(SItem.java:1663)
	at legend.game.inventory.screens.UseItemScreen.renderUseItemMenu(UseItemScreen.java:165)
	at legend.game.inventory.screens.UseItemScreen.render(UseItemScreen.java:105)
	at legend.game.inventory.screens.MenuScreen.renderScreen(MenuScreen.java:72)
	at legend.game.inventory.screens.MenuStack.propagate(MenuStack.java:128)
	at legend.game.inventory.screens.MenuStack.propagate(MenuStack.java:124)
	at legend.game.inventory.screens.MenuStack.render(MenuStack.java:100)
	at legend.game.Scus94491BpeSegment_8002.loadAndRenderMenus(Scus94491BpeSegment_8002.java:810)
	at legend.game.Scus94491BpeSegment.lambda$gameLoop$46(Scus94491BpeSegment.java:416)
	at legend.core.RenderEngine.lambda$init$18(RenderEngine.java:609)
	at legend.core.opengl.Window$Events.onDraw(Window.java:792)
	at legend.core.opengl.Window.tickFrame(Window.java:432)
	at legend.core.opengl.Action.run(Action.java:46)
	at legend.core.opengl.Action.tick(Action.java:32)
	at legend.core.opengl.Window.run(Window.java:408)
	at legend.core.RenderEngine.run(RenderEngine.java:1008)
	at legend.core.GameEngine.start(GameEngine.java:256)
	at legend.game.Main.main(Main.java:40)
```

And two other things to consider:
- Title screen doesn't support repeat input, should we add it there?
- Dropdown listbox supports wrapping but doesn't stop on repeats

warblgarbl:
Looking it over now, the title screen will take a bit more of an effort to get the repeat press working smoothly. As it is, if you just swap out the `pressedThisFrame` for `pressedWithRepeat` it will stop and restart.

Dropdown lists don't have any menu sounds currently. Should I add those in? If so, all three: highlight, select, cancel?

LordMonoxide (approved):

PR #2280, opened 2025-03-20
fix kongol I and dd cutscene asset loading race conditions

avionanx:


LordMonoxide (changes requested):
Needs comments explaining your changes. You can use # in the diffs

LordMonoxide (approved):

PR #2285, opened 2025-03-28
Floaterized battle speed addition

Illeprih:
This should still match retail speed calculations, but it should allow easier implementation of actual rand()


LordMonoxide:
Thanks for the proofs, I converted to floats and removed the bithacking constants

PR #2310, opened 2025-06-06
Fix loading from menu

### LINKED ISSUE #2225: Load from Main Menu: Aglis Psyche Bomb Trial Missing NPCs
Savan and Lulu don't load into the correct position. Same result when just warping to the submap with the debugger. Something particular about how the map is loaded from the title screen.

avionanx:
Got some fields to name later, _800bd7ac, _800bd7b0

### DIFF
--- a/src/main/java/legend/game/inventory/screens/MainMenuScreen.java
+++ b/src/main/java/legend/game/inventory/screens/MainMenuScreen.java
@@ -49,6 +49,7 @@ import static legend.game.Scus94491BpeSegment_8005.standingInSavePoint_8005a368;
 import static legend.game.Scus94491BpeSegment_8005.submapCutForSave_800cb450;
 import static legend.game.Scus94491BpeSegment_8005.submapCut_80052c30;
 import static legend.game.Scus94491BpeSegment_8005.submapScene_80052c34;
+import static legend.game.Scus94491BpeSegment_800b._800bd7ac;
 import static legend.game.Scus94491BpeSegment_800b.continentIndex_800bf0b0;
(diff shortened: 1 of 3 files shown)


avionanx:
Oops this messed WMAP

avionanx:
Needs to be tested with loading from disc change saves

src/main/java/legend/game/inventory/screens/MainMenuScreen.java
```diff
@@ -357,23 +361,34 @@ private void showLoadScreen() {
...
 
       submapScene_80052c34 = gameState_800babc8.submapScene_a4;
-      submapCut_80052c30 = gameState_800babc8.submapCut_a8;
+      if(gameState_800babc8.isOnWorldMap_4e4) {
+        submapCut_80052c30 = 0;
+      } else {
+        submapCut_80052c30 = gameState_800babc8.submapCut_a8;
+      }
+
+      if(currentEngineState_8004dd04 instanceof final SMap smap && !gameState_800babc8.isOnWorldMap_4e4) {
```
LordMonoxide, line 379: Can you explain these changes? The reason the engine state had loadGameFromMenu is so that the engine state gets to decide how to restore itself. What happens when a mod adds its own engine state that can save/load? We can't add another `else if instanceof` here for that

avionanx: Oops, forgot we already had custom engine states. Will re-add and move these

LordMonoxide (approved):

PR #2315, opened 2025-06-13
Moddable spell ring colours closes #2314

### LINKED ISSUE #2314: Moddable spell ring colours

Grey-Mad:


LordMonoxide:
This seems very heavy weight, what does a registry buy here? What is your use case?

Grey-Mad:
My assumption is that in the far future, when someone inevitably tries to add new elements and/or new dragoons, a way will be needed to add or reconfigure the existing spell rings. I based the changes off of the existing element registry because I am assuming the use cases would be very similar. If it is prudent to minimize the amount of registries added, would it be better idea to move spell ring colors (and weapon trail colors) into the existing element registry? If so, I think I could make those changes.

LordMonoxide:
This is part of the character template now

PR #2323, opened 2025-06-16
Discord Rich Presence Implementation

FrancisDionne:


src/main/java/discord/DiscordRichPresence.java
```diff
@@ -0,0 +1,163 @@
...
+    if(engineState_8004dd20 == EngineStateEnum.SUBMAP_05 && submapId_800bd808 > -1 && submapId_800bd808 < submapNames_8011c108.length) {
+      return submapNames_8011c108[submapId_800bd808];
+    }
+    if(continentIndex_800bf0b0 > -1 && continentIndex_800bf0b0 < worldMapNames_8011c1ec.length) {
+      return worldMapNames_8011c1ec[continentIndex_800bf0b0];
+    }
+    return "Unknown Location";
+  }
+
+  private static int getParty() {
```
LordMonoxide, line 151: Not sold on this. What other info can be put here?

src/main/java/discord/package-info.java
```diff
@@ -0,0 +1 @@
+package discord;
```
LordMonoxide, line 1: Needs package annotations

src/main/java/discord/DiscordRichPresence.java
```diff
@@ -0,0 +1,163 @@
...
+        core.setLogHook(LogLevel.VERBOSE, DiscordRichPresence::LogCallback);
+
+        startActivity();
+      } catch(final Exception ex) {
+        LOGGER.error(ex);
+        return;
+      }
+
+      try {
+        while(true) {
```
LordMonoxide, line 52: Use a boolean for this loop and set it to false to stop running. You can still interrupt the thread so that it stops sleeping.

LordMonoxide (changes requested):
Some CS issues here as well. IntelliJ style and inspections are included in the project.

src/main/java/discord/DiscordRichPresence.java
```diff
@@ -0,0 +1,163 @@
...
+  private static Activity activity;
+
+  private DiscordRichPresence() {
+  }
+
+  public static void start() {
+    thread = new Thread(() -> {
+      try {
+        final CreateParams params = new CreateParams();
+        params.setClientID(1383897032212611112L);
```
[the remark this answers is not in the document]
FrancisDionne: This is the ID for the Discord App. I think Drew mentioned you guys have a Discord dev group. We'd need to make an app there and use that ID instead.

src/main/java/discord/DiscordRichPresence.java
FrancisDionne: It should it theory never reach it. And if it does we would have to get it reported so we check why/where it happens. But it's not like we can do much in the little activity space it has. Opened to anything, even empty if you want.

src/main/java/discord/package-info.java
FrancisDionne: I'm new to this whole java thing so I'll need to look into that

src/main/java/discord/DiscordRichPresence.java
FrancisDionne: Roger roger

src/main/java/discord/DiscordRichPresence.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  private static void startActivity() {
+    activity = new Activity();
+    activity.timestamps().setStart(Instant.now());
+    activity.assets().setLargeImage("https://assets-prd.ignimgs.com/2022/03/02/the-legend-of-dragoon-1646184664725.jpg");
+  }
+
+  public static void updateActivity() {
+    if(gameState_800babc8 != null) {
```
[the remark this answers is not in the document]
FrancisDionne: I'm not sure I understand what the required change is here.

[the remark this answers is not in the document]
FrancisDionne: I'm not sure I get exactly what you require here. I think I get it but I wouldn't really know how/where to start.

src/main/java/discord/DiscordRichPresence.java
```diff
@@ -0,0 +1,163 @@
...
+      LOGGER.warn(message);
+    } else if(level == LogLevel.INFO) {
+      LOGGER.info(message);
+    }
+  }
+
+  private static void startActivity() {
+    activity = new Activity();
+    activity.timestamps().setStart(Instant.now());
+    activity.assets().setLargeImage("https://assets-prd.ignimgs.com/2022/03/02/the-legend-of-dragoon-1646184664725.jpg");
```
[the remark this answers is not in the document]
FrancisDionne: Drew hosted the picture and we're now using it

src/main/java/discord/DiscordRichPresence.java
LordMonoxide: Yeah, set to false and then interrupt. You don't have to handle the interruptedexception, the boolean being false would cause it to exit on the next loop. This also means that if the thread gets interrupted for any other reason, it wouldn't terminate, because the loop var would still be true.

PR #2425, opened 2025-09-26
Expand to include NixP

weegs710:


weegs710:
I have changed the approach. I created launch-nix to be it's own launch script and restored launch-linux and download-java to the way they were before I modified them. This should work much better as it no longer forces logic that would otherwise be unnecessary and no longer tramples on your codebase, and it is no longer pinned to any hashes so it should be much more future proof as well. Apologies.

PR #2492, opened 2025-12-17
implement addsp

avionanx:


src/main/java/legend/lodmod/items/RecoverSpItem.java
```diff
@@ -27,8 +29,19 @@ public int getMaxStackSize(final ItemStack stack) {
   }
 
   @Override
-  public boolean canBeUsed(final ItemStack stack, final UsageLocation location) {
-    return location == UsageLocation.BATTLE;
+  public boolean canBeUsed(final ItemStack stack, final UsageLocation location) { return true; }
```
LordMonoxide:
```suggestion
  public boolean canBeUsed(final ItemStack stack, final UsageLocation location) {
    return true;
  }
```

src/main/java/legend/lodmod/items/RecoverSpItem.java
```diff
@@ -27,8 +29,19 @@ public int getMaxStackSize(final ItemStack stack) {
...
+
+  @Override
+  public boolean canBeUsedNow(final ItemStack stack, final UsageLocation location) {
```
LordMonoxide: Remove the bool, return true in the loop instead of breaking, and return false in the last line

src/main/java/legend/game/SItem.java
```diff
@@ -627,8 +627,30 @@ public static int addMp(final int charIndex, final int amount) {
...
+    final int maxSp = stats.dlevel_0f * 100;
+    if(stats.sp_08 == maxSp) {
+      return -2;
+    }
+
+    final int spToAdd = amount == -1 ? maxSp - stats.sp_08 : Math.min(amount, maxSp - stats.sp_08);
+    final int responseType = amount == -1 || spToAdd < amount ? -1 : amount;
+
+    charData.sp_0c += spToAdd;
+    gameState_800babc8.charData_32c[charIndex].dlevelXp_0e += spToAdd;
```
LordMonoxide: We should probably add the whole amount to the xp eh?

src/main/java/legend/game/SItem.java
```diff
@@ -627,8 +627,30 @@ public static int addMp(final int charIndex, final int amount) {
 
   @Method(0x80022cd0L)
   public static int addSp(final int charIndex, final int amount) {
-    assert false;
-    return 0;
+    final CharacterData2c charData = gameState_800babc8.charData_32c[charIndex];
+    final ActiveStatsa0 stats = stats_800be5f8[charIndex];
+
+    final int maxSp = stats.dlevel_0f * 100;
+    if(stats.sp_08 == maxSp) {
```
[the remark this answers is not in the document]
avionanx: This returns "Nothing Happened" message, which is no longer used in SC in a way because of canBeUsedNow usages, it used to consume the item and not give you anything.

src/main/java/legend/game/SItem.java
avionanx: unless its -1

PR #2549, opened 2026-02-01
Blacklisted VOCG maps from saveanywhere

### LINKED ISSUE #2548: Saves on Valley of Corrupted Gravity often cause issues with floating rocks - blacklist saveanywhere
VoCG needs to be blacklisted for saveanywhere as it is often causing rocks to disappear or bug out, trapping the player and requiring them to debug teleport to another map.

samsonjm:
Fixes #2548


samsonjm:
I left out maps 251 and 258 (the gate), assuming that since there are no moving rocks there they would be okay.

PR #2553, opened 2026-02-05
Unix Environment variable allows for automatic choice for GPU selecti…

Dejin1:
**Original Issue:**

Users with laptops or systems that have both integrated graphics and discrete GPUs couldn't easily choose which GPU to use. The game would often default to the weaker integrated GPU, causing performance issues.

**What Was Done:**

Added launch.conf configuration file to control GPU preference (auto/discrete/integrated)
Linux/Steam Deck: Automatic GPU switching using the DRI_PRIME environment variable 

Just set your preference in launch.conf and launch Windows: Created optional gpu-optional-setup.bat helper script that configures Windows registry to respect GPU preference
Updated all platform launch scripts to support the new configuration
Added comprehensive documentation in README with both automatic and manual setup instructions

**Simple:**
If you have a laptop with two graphics cards, you can now tell Severed Chains which one to use. Linux users just edit a config file and it works automatically. Windows users can run an optional setup script that makes Windows remember your GPU choice. This helps ensure the game runs on your powerful GPU instead of the power saving one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -168,6 +168,7 @@ tasks.register('copySupportFiles', Copy) {
         from('download-java') {
           filter(ReplaceTokens, tokens: [os: 'linux', arch: isArm ? "aarch64" : "x64"])
         }
+        from 'launch.conf'
         break
 
       case OperatingSystem.MAC_OS:
@@ -183,6 +184,7 @@ tasks.register('copySupportFiles', Copy) {
         from('download-java') {
           filter(ReplaceTokens, tokens: [os: 'macosx', arch: isArm ? "aarch64" : "x64"])
         }
+        from 'launch.conf'
         break
 
       case OperatingSystem.WINDOWS:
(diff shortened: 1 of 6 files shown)


launch.conf.example
```diff
@@ -0,0 +1,17 @@
+# Severed Chains Launch Configuration
```
LordMonoxide: Since the launch script copies this to `launch.conf` on run if it doesn't exist and sources it unconditionally, let's just rename this file to `launch.conf` and remove the copy code from the launch scripts.

build.gradle
```diff
@@ -194,9 +194,11 @@ tasks.register('copySupportFiles', Copy) {
           filter(ReplaceTokens, tokens: [false: "true"])
           filter(ReplaceTokens, tokens: [libs: libs])
         }
+        from 'gpu-optional-setup.bat'
         break
     }
 
+    from 'launch.conf.example'
```
LordMonoxide: This file is only needed on linux and mac, so move it up into their switch cases

PR #2571, opened 2026-02-17
Auto update system

Dejin1:
Added a standalone updater JAR that downloads release archives from github extracts and applies new files.

### Prerequisites
- A built copy of Severed Chains (any previous release)
- The built `updater.jar` from this branch (`build/libs/updater.jar`)

### Steps

1. Build the updater:
   ```
   set JAVA_HOME=jdk25.0.0_36
   gradlew updaterJar
   ```

2. Copy `build/libs/updater.jar` into an existing Severed Chains install directory.

3. Get the latest release asset URL:
   ```
   curl -s https://api.github.com/repos/Legend-of-Dragoon-Modding/Severed-Chains/releases | jq -r '.[0].assets[] | select(.name | test("Windows")) | .browser_download_url'
   ```

4. Run the updater directly (no game process needed for testing):
   ```
   jdk25.0.0_36\bin\java -jar updater.jar <URL_FROM_STEP_3> <GAME_DIR>
   ```

5. Verify:
   - New game JAR appears, old one is removed
   - `libs/` contains only current-version JARs
   - `update_log.txt` is created with the update summary
   - `saves/`, `mods/`, `isos/`, `files/` are untouched
   - Game relaunches automatically
   
------------------------------------------------

### Testing In-Game Auto-Update

To test the full user experience with the title screen button:

1. Launch an this build of Severed Chains and make sure there is an update available.

2. Wait for the title screen. If an update is available and GitHub has a platform-specific release asset, the "Update Available" button will appear.

3. Click the "Update Available" button. The game will:
   - Launch `updater.jar` as a separate process
   - Exit immediately (so files can be overwritten)

4. The updater console window will appear showing progress:
   ```
   [Updater] Waiting for game process to exit...
   [Updater] Downloading update from https://github.com/.../Severed_Chains_Windows.zip
   [Updater] Download complete.
   [Updater] Extracting update archive...
   [Updater] Applying update files...
   [Updater] Update complete. Please restart Severed Chains.
   ```

5. After the updater finishes, manually restart the game using the normal launch script (`launch.bat`, `launch-linux`, etc.).

6. Verify:
   - Game launches with the new version
   - `update_log.txt` contains the update summary
   - All user data intact

### Test Proof (February 15, 2026)

https://cdn.discordapp.com/attachments/1055652902720516126/1472831956285657148/Patched.mp4?ex=6994a9e6&is=69935866&hm=e07ee90f3c8877210bd90f87fa584ed82e20119fd7cb31d040054aa0eb25004e&


src/main/java/legend/core/SelfUpdater.java
```diff
@@ -0,0 +1,135 @@
...
+      LOGGER.error("Failed to launch updater", e);
+      progressCallback.accept(new UpdateProgress(UpdateState.FAILED, "Failed to launch updater: " + e.getMessage()));
+    }
+  }
+
+  /**
+   * finds the Java executable preferring the bundled JDK.
+   */
+  private static Path findJavaExecutable(final Path gameDir) {
+    final String os = System.getProperty("os.name", "").toLowerCase(Locale.US);
```
LordMonoxide, line 112: Look into `String jvmPath = ProcessHandle.current().info().command()`, this should give the path to the currently running jvm so imo that should be our first step in this method

src/main/java/legend/core/SelfUpdater.java
```diff
@@ -0,0 +1,135 @@
...
+        Version.FULL_VERSION
+      );
+      pb.directory(gameDir.toFile());
+      pb.inheritIO();
+      pb.start();
+
+      progressCallback.accept(new UpdateProgress(UpdateState.DONE, "Updater launched. Game will close now."));
+
+      // give the updater a moment to start
+      try { Thread.sleep(1000); } catch(final InterruptedException ignored) { }
```
LordMonoxide: Do we need to wait here? This smells like a race condition. If starting the process is synchronous I don't see a reason to wait, and if it's async then what happens if it takes longer than one second to start?

src/main/java/legend/core/SelfUpdater.java
```diff
@@ -0,0 +1,135 @@
...
+
+      // exit the game so the updater can overwrite files
+      LOGGER.info("Exiting game for update...");
+      System.exit(0);
```
LordMonoxide: I'd prefer to shut down the normal way `RENDERER.window().close()`. This won't shut down the native libraries properly and will cause some driver APIs to segfault or otherwise crash the jvm

.gitignore
```diff
@@ -36,6 +36,7 @@ hs_err_pid*
 .gradle
 /build/
 /out/
+.idea/
```
LordMonoxide: I have a separate .gitignore in this directory

src/main/java/legend/core/UpdaterMain.java
```diff
@@ -0,0 +1,716 @@
...
+  /** Directories and files that should never be overwritten by an update */
+  private static final Set<String> PRESERVED = Set.of(
+    "saves",
+    "mods",
+    "isos",
+    "files",
+    "jdk25",
+    "config.dcnf",
+    "launch.conf",
+    "updater.jar"
```
LordMonoxide, line 60: What happens if we update the updater?

src/main/java/legend/core/UpdaterMain.java
```diff
@@ -0,0 +1,716 @@
...
+   */
+  private static void waitForProcessExit(final long pid, final long timeoutMs) throws InterruptedException {
+    final long deadline = System.currentTimeMillis() + timeoutMs;
+
+    while(System.currentTimeMillis() < deadline) {
+      final var optProcess = ProcessHandle.of(pid);
+      if(optProcess.isEmpty() || !optProcess.get().isAlive()) {
+        System.out.println("[Updater] Game process has exited.");
+        // Brief extra pause to let file handles release
+        Thread.sleep(1000);
```
LordMonoxide: Is this necessary? Once the jvm process exists, all of our resources should already be free, shouldn't they?

src/main/java/legend/core/UpdaterMain.java
Dejin1: instead of overwriting the running jar we can stage an incoming updater.jar as updater.jar.new and have the game swap it in at the top of Main.main() on next launch...maybe? when nothing holds the lock. we can keep the updater's CLI args backward compatible since the previous updater always drives the next release.

src/main/java/legend/core/UpdaterMain.java
Dejin1: On my machine there was a few times where this was a bit glitchy to not have this small wait time but i can remove it and see if others see this issue or not it was flaky when it happened so maybe it wont rear up again.

src/main/java/legend/core/SelfUpdater.java
Dejin1: fixed in next commit

src/main/java/legend/core/SelfUpdater.java
Dejin1: fixed in next commit

src/main/java/legend/core/SelfUpdater.java
Dejin1: fixed in next commit

.gitignore
Dejin1: This may have accidentally got in apologies

src/main/java/legend/core/UpdaterMain.java
Dejin1: with the staged swap now in place updater.jar.new TO updater.jar on next boot the updater can self update going forward

LordMonoxide (approved):

Issue #2527, opened 2026-01-10
Incorrect SP #'s in menu


Popoto-anonymous:
Haschel's 5-Ring Shattering and Meru's Hammer Spin display incorrect SP earned #'s in the Additions menu and the in combat addition swap menu. Attached are images taken via phone of PS5 Classic version (closest to retail I have) and Severed Chains version 3.0.0-1487-devbuild respectively

![Image](https://github.com/user-attachments/assets/a5530d4b-c9be-45f1-a53f-c19b3eae65a0)

<img width="2035" height="1570" alt="Image" src="https://github.com/user-attachments/assets/5d0a2ddf-7ec4-4894-b2ce-e6ba3554c24c" />

![Image](https://github.com/user-attachments/assets/1be21e08-7bba-4ecf-97d2-056587445531)

<img width="2043" height="1570" alt="Image" src="https://github.com/user-attachments/assets/61887d26-509d-45d0-b7c4-c465e1868db0" />

PR #2573, opened 2026-02-20
  Fix incorrect SP values for 5-Ring Shattering and Hammer Spin (clos…

Kunaixfox:
…es #2527)

  When additions were migrated from pre-computed SP lookup tables to
  multiplier-based calculation in the addition registry (#2475), the SP
  multipliers for these two additions were set to values that don't
  reproduce the original PS1 data (additionData_80052884).

  5-Ring Shattering SP multipliers:
    - LV 2: 1.10 -> 1.00 (was showing 38, should be 35)
    - LV 3: 1.20 -> 1.15 (was showing 42, should be 40)

  Hammer Spin SP multipliers:
    - LV 3: 1.50 -> 1.46 (was showing 52, should be 51)
    - LV 4: 1.75 -> 1.69 (was showing 61, should be 59)


Kunaixfox:
the following additions also have SP mismatches vs the original
  PS1 lookup table at certain levels, caused by the same multiplier
  rounding issue:
    - Dart - Crush Dance: LV 2 (+2), LV 4 (+2)
    - Lavitz - Rod Typhoon: LV 5 (+2)
    - Lavitz - Flower Storm: LV 5 (+2)
    - Meru - Double Smack: LV 5 (+1)
    - Meru - Cool Boogie: LV 5 (+4)
    - Haschel - Summon 4 Gods: LV 2 (+1), LV 4 (+1), LV 5 (+1)
    - Albert - Rod Typhoon: LV 5 (+2)
    - Albert - Flower Storm: LV 5 (+2)
[Full_verify_sp_fix.py](https://github.com/user-attachments/files/25443808/Full_verify_sp_fix.py)
if you would like to open an issue after double check these i can also correct them.

PR #2589, opened 2026-02-27
SC Credits/AboutScreen

FrancisDionne:
As discussed, a Severed Chains credits screen accessible from the ex LinksScreen now renamed AboutScreen. 

The LOD credits are also accessible from there. Just be aware that there will be something to do there as they currently crash because of the audio issue. Could hide that line for now if that's not in your priorities.


src/main/java/legend/game/inventory/screens/CreditsScreen.java
```diff
@@ -0,0 +1,350 @@
...
+
+  private int loadingStage;
+  private final Runnable unload;
+
+  private final Matrix4f m;
+  private final Obj quad;
+  private final Texture[] textures;
+
+  private final List<CreditEntry> credits;
+  private final HashMap<CreditsType, CreditFontProperties> fonts;
```
LordMonoxide: It's best practice to use the base class for collection vars so that the implementation can be swapped out without having to change the declaration
```suggestion
  private final Map<CreditsType, CreditFontProperties> fonts;
```

LordMonoxide (changes requested):

PR #2660, opened 2026-04-23
Fix for menu addition values (closes #2659)

### LINKED ISSUE #2659: Omni-Sweep and Cat's Cradle Damage in Addition Menu Differing from Retail
In retail the damage is 501% for Omni-Sweep but in SC it is showing as 500%. Also happening for Cat's Cradle which normally deals 351% but appears as 350% here.

Retail:
Image

SC:
Image

Retail: 
Image

SC:
Image

Illeprih:
The issue is caused by float math.
There are multiple += hit * multi, since the value is floored at the end, this compounds the float imprecision enough to shift it slightly below the actual value.

### DIFF
--- a/src/main/java/legend/game/additions/SimpleAddition.java
+++ b/src/main/java/legend/game/additions/SimpleAddition.java
@@ -18,10 +18,10 @@ public abstract class SimpleAddition extends Addition {
     float damage = 0;
 
     for(int hit = 0; hit < this.hits.length; hit++) {
-      damage += this.hits[hit].damageMultiplier_04 * multi;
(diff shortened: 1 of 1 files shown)


src/main/java/legend/game/additions/SimpleAddition.java
```diff
@@ -18,10 +18,10 @@ public int getDamage(final GameState52c state, final CharacterData2c charData, f
     float damage = 0;
```
LordMonoxide: Should change this to `int` since we're just accumulating int values now

LordMonoxide (approved):

PR #2671, opened 2026-05-02
Bugfix: Add event emission to level up events

pkolb-dev:
Currently, the following events do not emit after the char branch was merged:
- CharacterDragoonLevelUpEvent
- CharacterLevelUpEvent

This PR adds the event emission to the appropriate methods.

### DIFF
--- a/src/main/java/legend/lodmod/characters/RetailCharacterTemplate.java
+++ b/src/main/java/legend/lodmod/characters/RetailCharacterTemplate.java
@@ -185,7 +185,7 @@ public abstract class RetailCharacterTemplate extends CharacterTemplate {
 
   @Override
   public void applyLevelUp(final CharacterData2c character, @Nullable final LevelUpActions actions) {
-    CharacterLevelUpEvent event = new CharacterLevelUpEvent(character);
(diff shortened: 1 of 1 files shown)


src/main/java/legend/lodmod/characters/RetailCharacterTemplate.java
```diff
@@ -193,6 +194,8 @@ public void applyLevelUp(final CharacterData2c character, @Nullable final LevelU
     event.statsToAdd.put(MAGIC_DEFENSE_STAT.get(), this.getMagicDefenseToAdd(character.level_12));
     event.statsToAdd.put(SPEED_STAT.get(), this.getSpeedToAdd(character.level_12));
 
+    event = EVENTS.postEvent(event);
```
LordMonoxide: This returns what was passed in for convenience, so you can remove the `event = ` part and add the final back to the variable. Same for the dragoon one. Other than that it looks good.

pkolb-dev: I think I understand up to the `add the final back to the variable`. i've made a commit, please review

LordMonoxide (approved):

PR #2675, opened 2026-05-03
Move Dragoon sp bar colours to character template closes #2670

### LINKED ISSUE #2670: Dragoon SP Colours
Move Dragoon SP bar colours to the character template so they can be overriden with Mods.

Zychronix:


src/main/java/legend/lodmod/characters/RetailCharacterTemplate.java
```diff
@@ -357,4 +358,9 @@ protected void swapSpell(final CharacterData2c character, final RegistryId from,
       character.addSpell(to, info);
     }
   }
+
+  @Override
+  public int[][] getDragoonSpBar() {
```
LordMonoxide: Small change, make this take a `level` param and only return one level at a time, and rename to `getSpBarColours`

Zychronix: Done!

LordMonoxide (approved):

PR #2684, opened 2026-05-11
fix simplerand spell status chances (closes #2440)

### LINKED ISSUE #2440: Imago Status Attack Chance
In SC: Imago's dispiriting attack inflicted dispiriting 20 times in a row. I've been operating under the assumption that status effects from enemies are either 50% or 100% chance to be inflicted, so either this is 100% or if it is 50%, then I landed on tails 20 times in a row which is a 1 in 1,048,576 chance. Literally about 1 in a million.

In retail: I confirmed that it can fail to proc, albeit I just went 15 times in a row before it failed to inflict. If that was 50%, then it would be 1 in 32,768 odds. 

I have already eliminated a-av and m-av (all party members had 0% for both,) and gear was equipped that did not grant immunity (which I doubly would have noticed as the character who resisted the status had been afflicted multiple times earlier in the same battle.)

My best guess is that the base chance is neither 50% nor 100%, or that there is a bug in retail that sometimes causes the status to fail and it has been patched in SC.

Image

avionanx:


avionanx:
whoops I imported wrong package lol

avionanx:
nevermind I need to sleep

LordMonoxide (approved):

PR #2685, opened 2026-05-12
Adjust Dragoon lvl and Character level events

pkolb-dev:
- rename existing events to use "pre" prefix
- create new events for "post" events for purely informational handling.


src/main/java/legend/game/modding/events/characters/PostCharacterDragoonLevelUpEvent.java
```diff
@@ -0,0 +1,16 @@
+package legend.game.modding.events.characters;
+
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
```
LordMonoxide: Looks like several dead imports here

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #2694, opened 2026-05-31
Fix 0 byte updater.jar in distribution builds

Dejin1:
Fixes the overwrite of the updater file by placing it in a different dir


LordMonoxide:
Thanks for the PR, found a cleaner way to fix it in 7598ca69

PR #2711, opened 2026-06-27
CRT postprocessing enhancements

sjbuck:
- Multiple resolution options for Pixelate effect
- Add bloom effect to simulate bright pixels bleeding into neighbors
- Gamma-correct grille effect to minimize dimming
- Add bilinear filtering to improve chromatic aberration appearance
- Snap ortho rendering to integer pixels when Pixelate is enabled
- Fudge low-resolution pixelation on Turn Order UI to maintain legibility


sjbuck:
Severed Chains 3 0 0-SNAPSHOT-CHANNEL - FPS_ 60 00_60 scale_ 8 00 res_ 368x240 6_26_2026 10_59_45 AM

TODO: Add an option to use the original pre-baked battle text sprites (i.e. character names, "ATTACK", "GUARD", etc.) for pixel-perfect fidelity to the original. Maybe also switch to the 8px font from `font.tim` for the Turn Order UI, since the main font is still a bit fuzzy at that small size, even with the resolution fudging I added.

src/main/java/legend/core/RenderBatch.java
```diff
@@ -251,7 +262,16 @@ public void updateProjections() {
 
     this.needsSorting = true;
     final T entry = this.orthoPool.acquire(type);
-    entry.acquire(obj, this.sequence++, transforms);
+    final boolean snap = CONFIG.getConfig(CoreMod.SHADER_ENABLE_CRT_CONFIG.get()) && CONFIG.getConfig(CoreMod.SHADER_PIXELATE_CONFIG.get()) != legend.core.opengl.PixelateMode.NONE;
+    if (snap) {
+      this.temp.set(transforms);
+      final Vector3f translation = new Vector3f();
```
LordMonoxide: Java doesn't natively support stack allocation for objects, so anything that is `new`ed gets allocated on the heap, and needs to be garbage-collected. This is a problem in code called by the render loop, especially methods that are called potentially many times.

Each instance of this object will have a 12-byte header and 3 4-byte floats so for example, if we were to queue 100 ortho objects (not uncommon with text on the screen) at 60 FPS, this would allocate 24 * 100 * 60 = ~140kb per second, which adds up very quickly.

To work around this limitation we usually use a private class variable (like `this.temp`), but in this case you can read the translation directly from the matrix:
```java
this.temp.setTranslation(Math.round(transforms.m30()), Math.round(transforms.m31()), Math.round(transforms.m32()));
```

I might also suggest adding a new private method to this class that checks if it should snap and if so rounds the matrix's translation.

src/main/java/legend/turnorder/TurnOrderMod.java
```diff
@@ -74,11 +76,13 @@ public void onGameLoaded(final GameLoadedEvent event) {
   @EventListener
   public void battleEndedEvent(final BattleEndedEvent event) {
     this.visible = false;
+    isVisible = false;
```
LordMonoxide: CS: `this.` before all class members (applies below too)

src/main/java/legend/core/RenderEngine.java
```diff
@@ -759,8 +767,42 @@ public void init() {
...
           screenShaderOptions.scanlinesOpacity(CONFIG.getConfig(SHADER_SCANLINES_OPACITY_CONFIG.get()));
           screenShaderOptions.grilleOpacity(CONFIG.getConfig(SHADER_GRILLE_OPACITY_CONFIG.get()));
-          screenShaderOptions.resolution(this.window.getWidth() / 2.0f, this.window.getHeight() / 1.5f);
-          screenShaderOptions.pixelate(CONFIG.getConfig(SHADER_PIXELATE_CONFIG.get()));
+          final PixelateMode pixelateMode = CONFIG.getConfig(SHADER_PIXELATE_CONFIG.get());
+          screenShaderOptions.pixelate(pixelateMode != PixelateMode.NONE);
+
+          final float resWidth;
+          final float resHeight;
+          switch (pixelateMode) {
```
LordMonoxide: CS: no spaces after control statements or casts (applies elsewhere as well)

Our IntelliJ code style is included in the repo - if you highlight your code and press `ctrl+alt+L` it will format it automatically

LordMonoxide (changes requested):

LordMonoxide (approved):

PR #2751, opened 2026-08-03
Single monster target event closes #2667

### LINKED ISSUE #2667: Monster Targeting Events
New events that show which monster the player is targeting.

Zychronix:


src/main/java/legend/game/combat/ui/BattleHud.java
```diff
@@ -776,6 +764,7 @@ private void drawUiElements() {
             //LAB_800f0d10
             str = this.getTargetEnemyName(monsterBent, this.battle.currentEnemyNames_800c69d0[enemySlot]);
             targetBent = monsterBent;
+            EVENTS.postEvent(new SingleMonsterTargetEvent(this.battle, monsterBent));
```
LordMonoxide, line 767: This fires every frame, surely we can move it to where target selection happens so it only fires when the target is changed?

LordMonoxide (changes requested):

Zychronix:
PR not needed anymore. RenderEvent is sufficient.

PR #2754, opened 2026-08-03
Equip effects in Dragoon form closes #2300

### LINKED ISSUE #2300: Combat script modding events

Zychronix:


patches/scripts/player_combat_script.txt
```diff
@@ -6512,9 +6512,12 @@ consume
 
 ; SUBROUTINE
 HANDLE_REGEN:
+call Scus94491BpeSegment::scriptReadRegistryEntryVar, id[lod_core:config], id[lod_core:equip_effects_in_dragoon], 0, stor[31]
+jmp_cmp ==, 0x1, stor[31], inl[:HANDLE_REGEN_1] ;
```
LordMonoxide: In C-like languages, true is generally defined as "non-zero", it's safer to make both of these conditions `!=, 0` instead of `==, 1`

LordMonoxide (changes requested):

patches/scripts/player_combat_script.txt
Zychronix: Update to use !=, 0

LordMonoxide (approved):

PR #2766, opened 2026-08-14
BattleStage Leak Fix

dragoonsouls:
Battle Stage leaks shown when deleting the GL Objects after deallocating the Battle, now* seems to be solved, extracted the deleting code from deallocateCombat(), create a new Method called deleteBattleStageModel() and set a call inside deallocateCombat() and loadStageTmdAndAnim(), so each time a Battle starts or end, first will check if a model is loaded, then in the replacement will ensure to delete all the GL corresponding to the old BattleStage.


LordMonoxide (approved):

LordMonoxide:
#790

PR #2767, opened 2026-08-15
Remember Campaign Settings

Ink230:
## Motivation

Campaign-specific settings previously reset to registered defaults whenever a new campaign was created. This was super annoying as 95% of my campaign preferences were always the same.

## Changes

- Adds a global toggle that remembers campaign settings for future campaigns
- Stores dynamic, type-checked snapshots while safely ignoring removed, renamed, incompatible, external-mod, or malformed entries
- Adds campaign configuration lifecycle events so mods can persist their settings and override restored defaults
  - The system does not save any mod stuff on its own and leaves that up to modders to deal with
  - Modders get an end of lifecycle snapshot and can overwrite any settings they want

## Other bugs fixed

- Restores remembered controls into the live input-binding table when creating a new campaign
- Resolves saved input actions against the current registry after mod reboot, fixing controls when loading existing campaigns (and new game transitions)
  - Have been wondering why control edits were not persisting into a new campaign load 

## QA
- https://streamable.com/d83v5v

## Notes
- Existing or previous campaigns do not get updated (could be a feature though: copy existing sticky config to current campain)
- Campaign saves are still authoritative for campaign loads
- Remember Campaign Config setting is on by default but can be turned off

### DIFF
--- a/src/main/java/legend/core/GameEngine.java
+++ b/src/main/java/legend/core/GameEngine.java
@@ -339,6 +339,7 @@ public final class GameEngine {
 
     // Load default bindings for input actions
     InputBindings.initBindings();
+    InputBindings.loadBindings(CONFIG);
 
     loadLang();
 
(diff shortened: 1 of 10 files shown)


Ink230:
Presets added instead: #2804

PR #2774, opened 2026-08-18
Isolate held axes across controllers

### LINKED ISSUE #2193: Stick drift on idle controller interferes with input on active controller
When an axis is in-use on a controller and an idle controller is experiencing drift on that same axis, the drift input is registered as a release, interrupting any held input on the active controller.

Ink230:
Fixes #2193

### Motivation

Current axis input handling is under baked with some gaps in handling capabilities. This becomes apparent with 2193 or multiple controllers. Fixes scope larger than just 2193 but that is the immediate impact.

### Desc
- Isolated per controller axis tracking
  - Identifies held input by controller, axis, and direction so drift from another controller cannot release the active controller’s input
- Action value aggregation
  - Combines active axis sources and uses the strongest value, releasing the action only after every contributing source becomes inactive
- Lifecycle cleanup
  - Removes tracked axis input when a controller disconnects, the window loses focus, or action state resets, preventing stuck inputs
- Steam Input tracking
  - Keys held-axis suppression by controller and axis, preventing one controller’s drift from clearing another controller’s held marker

### QA
- Two controllers plugged in, only one is controlling the axis inputs despite the second controller joystick moving around
- Plug and play still working
  - Controller still works after plug and play.
```
23:23:38.451 [Hardware legend.core.platform.SdlPlatformManager:494] INFO : Gamepad XInput Controller disconnected
23:23:41.972 [Hardware legend.core.platform.SdlPlatformManager:483] INFO : Gamepad XInput Controller connected
```
- Input events fire with no external API surface changes

### DIFF
--- a/src/main/java/legend/core/platform/SdlPlatformManager.java
+++ b/src/main/java/legend/core/platform/SdlPlatformManager.java
@@ -4,8 +4,13 @@ import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.ints.IntOpenHashSet;
 import it.unimi.dsi.fastutil.ints.IntSet;
+import it.unimi.dsi.fastutil.longs.Long2FloatMap;
+import it.unimi.dsi.fastutil.longs.Long2FloatOpenHashMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.longs.LongIterator;
+import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
+import it.unimi.dsi.fastutil.longs.LongSet;
 import legend.core.platform.input.AxisInputActivation;
 import legend.core.platform.input.ButtonInputActivation;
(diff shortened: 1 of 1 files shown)


LordMonoxide:
This looks good but I have one question
> Combines active axis sources and uses the strongest value, releasing the action only after every contributing source becomes inactive

Does this mean that the InputAciton will only be released once all controllers (including a drifting controller) is released?

I guess if it's drifting, it would be under the deadzone, so its input won't contribute at all?

Ink230:
'all controllers' ->

- All connected controllers: no
- All connected controllers that have drift but are within the deadzone: no
- All connected controllers that drift outside of the deadzone: yes
  - Has to also be the same directional action
- All connected controllers with real input outside of deadzone: yes

This is in comparison to ALL of these being 'detected' as input before this fix.

For the last two points...does any game mitigate these two scenarios in any way...they are basically indistinguishable from each other.

Notes

- If a controller had input pending but was disconnected, it would also remain active to SC. This fix also removes its activity when the controller is disconnected
- You could assign a controller to be Player 1, and ignore all other controllers that are not on the Player 1 but do we want to manage controller player #s as well to mitigate drift? We would probably want to justify doing Players for a real reason like mods or co-op stuff
- Button presses set a controller to the 'active' controller as buttons to not drift but again weird use cases for people may hate this and turning a joystick into a button sounds like a nightmare for accessibility folks
 
🤷

PR #2777, opened 2026-08-20
Fix speed up/down buff issue closes #2776

### LINKED ISSUE #2776: Speed Up/Down incorrectly applying based off of base stats only.
New buff system only applies buffs based on based stats.
Example: Kongol in retail with 30 base speed and 20 speed bandit shoes should have 100 speed when Speed Up is used on him. In SC it is only 80.

Zychronix:


Zychronix:
The serializers for the mods needs to be compatible with the old and new versions.

PR #2778, opened 2026-08-21
Prevent timestamp from being set if unlockstate didn't change

pkolb-dev:
If the addition's unlockstate has not changed, we do not need to update the timestamp

### DIFF
--- a/src/main/java/legend/game/characters/CharacterAdditionInfo.java
+++ b/src/main/java/legend/game/characters/CharacterAdditionInfo.java
@@ -26,6 +26,10 @@ public class CharacterAdditionInfo {
   }
 
   public void setUnlockState(final UnlockState unlockState, final int timestamp) {
+    if (this.unlockState == unlockState) {
+      return;
+    }
+
     this.unlockState = unlockState;
 
     if(unlockState.isUsable()) {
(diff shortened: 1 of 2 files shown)


Ink230:
### Thoughts
1. I think the mod consuming events downstream of this change should be made more SC compliant and handle the innate SC behaviour better.
2. Any lifecycle stuff that influences unlock state should be handled by the mod consumer using lifecycle states from other events like in #2775 (or ones that need to be added for mods to figure out the lifecycle).
3. Timestamps and the unlock state work soundly in SC as is.

Maybe explore what the problem is more and we can come up with a better fix?

### Guards, local policy, global behaviour
This is not the right place for the guard. Not really the responsibility nor lifecycle location.

Any guard in here should also act on the timestamp parameter as well...but that doesn't make sense to do in any scenario I can think of. The change makes the method poorly named and basically implies it shouldn't be getting called from where it is currently being called from.

The guard also silently fails with your approach which is not ideal.

The timestamp and when something was unlocked is used to sort additions too. This change makes that unreliable.

A guard or check at the caller level is more appropriate. The debugger follows this pattern already.

```
final UnlockState targetState = this.additionUnlockState.getValue();

if(info.getUnlockState() != targetState) {
  info.setUnlockState(targetState, gameState.timestamp_a0);
}
```

We keep timestamp use cases untouched for saves, loading, transitions...etc.

I don't think there is any guard to add anywhere in SC for this.


### Incorrect Premise
The premise is also wrong

`If the addition's unlockstate has not changed, we do not need to update the timestamp`

does not align with

```
UNLOCKED, timestamp 120
UNLOCKED, timestamp 9000
```

These are not equivalent to SC, to the method being called, or the downstream effects. The method as is is facilitating this role correctly. Ordering, saves, loading...etc, lots of things rely on the key being <unlockstate, timestamp>.

So we do need to change the timestamp.

PR #2779, opened 2026-08-21
feature: add new AdditionLevelUpEvent to PostBattleScreen

pkolb-dev:
Right now archipelagoon has to wait until the next character level up to send the addition mastery unlock (blazing dynamo, flower storm, etc). this will allow mods to be notified when an addition levels up.


Ink230:
1. `PostAdditionLevelUpEvent` might be better naming and align with emerging conventions
2. `RegistryId additionId`, `CharacterAdditionInfo` should be considered to be added to the events payload

pkolb-dev:
>  1. `PostAdditionLevelUpEvent` might be better naming and align with emerging conventions
> 
>  2. `RegistryId additionId`, `CharacterAdditionInfo` should be considered to be added to the events payload

I agree with adding 1, so i'll do that shortly.

for 2, I'll adjust `Addition addition` to be `RegistryId additionId`. ~~What is the reasoning for adding CharacterAdditionInfo? or rather what information does it provide that is useful here out of curiosity?~~

I've swapped the `int level` to be `CharacterAdditionInfo additionInfo`

src/main/java/legend/game/modding/events/characters/PostAdditionLevelUpEvent.java
```diff
@@ -0,0 +1,18 @@
+package legend.game.modding.events.characters;
+
+import legend.game.characters.CharacterAdditionInfo;
+import legend.game.characters.CharacterData2c;
+import org.legendofdragoon.modloader.events.Event;
+import org.legendofdragoon.modloader.registries.RegistryId;
+
+public class PostAdditionLevelUpEvent extends Event {
```
LordMonoxide, line 13: Add javadoc on the class describing when it is fired and noting that additions may level up more than once in one fight so it may fire multiple times

LordMonoxide (changes requested):

src/main/java/legend/game/modding/events/characters/PostAdditionLevelUpEvent.java
pkolb-dev: Done. let me know if i need to add anything else to it

LordMonoxide (approved):

PR #2781, opened 2026-08-22
Decouple rendering from OpenGL

LordMonoxide:


Ink230:
image
 image

Funny guy.

LordMonoxide:
> image image
> 
> Funny guy.

Mostly updated imports :) tried to keep the usage the same, but I had to move classes around

src/main/java/legend/core/renderer/RenderEngine.java
```diff
@@ -1387,11 +1291,10 @@ private void resizeRenderBuffers() {
 
       this.renderTextures[i] = Texture.create("Render buffer " + i, builder -> {
         builder.size(this.renderWidth, this.renderHeight);
-        builder.internalFormat(GL_RGBA16);
-        builder.dataFormat(GL_RGBA);
-        builder.dataType(GL_UNSIGNED_BYTE);
-        builder.magFilter(GL_NEAREST);
-        builder.minFilter(GL_LINEAR);
+        builder.internalFormat(TextureInternalFormat.RGBA_8);
```
Ink230, line 1293: why gl_rgba16 to rgba_8

src/main/java/legend/core/renderer/RenderEngine.java
```diff
@@ -158,17 +105,18 @@ public class RenderEngine {
   private final List<RenderBatch> batches = new ArrayList<>();
   private final RenderBatch mainBatch;
   public final ScissorStack scissorStack;
-  private final RenderState state;
+  private RenderApi api = new GlesApi(this);
+//  private RenderApi api = new GlesApi(this);
```
Ink230: dupe comment blasphemy

src/main/java/legend/core/renderer/RenderEngine.java
LordMonoxide: We used to need the expanded bit depth for certain effects to work because certain layered effects weren't being calculated correctly, but that's no longer an issue and we can use standard 32-bpp textures for better performance on low-end devices.

PR #2788, opened 2026-08-27
Auto save post battle, create save from submap hotkey

Ink230:
### Motivation

At first I was angry that I had to open a menu to save. Then I became angry that I had to remember to save.

### Desc

- Uses all the existing guardrails to verify we can save
- Sound bank 50 sounds the best for save confirmation
  - On the fence but thinking its fine if there is no confirmation sound on success, just that on error there is index 40 played

### QA

image

image


src/main/java/legend/game/submap/SMap.java
```diff
@@ -435,6 +447,12 @@ public SMap() {
   public void init() {
     lastSavableEngineState = this.type;
     sssqResetStuff();
+    this.autoSaveAfterBattlePending = CONFIG.getConfig(CoreMod.SAVE_ANYWHERE_CONFIG.get())
+      && CONFIG.getConfig(CoreMod.AUTO_SAVE_AFTER_BATTLE_CONFIG.get())
+      && previousEngineState_8004dd28 == LodEngineStateTypes.BATTLE.get()
+      && postBattleAction_800bc974 != null
```
LordMonoxide: I don't think we need to check the postBattleAction here, even if the player runs away it wouldn't hurt to save?

src/main/java/legend/game/submap/SMap.java
```diff
@@ -541,10 +559,52 @@ public boolean canSave() {
...
+      return;
+    }
+
+    if(!this.canCreateNewSave()) {
+      this.autoSaveAfterBattleStableTicks = 0;
+      return;
+    }
+
+    this.autoSaveAfterBattleStableTicks++;
+    if(this.autoSaveAfterBattleStableTicks < AUTO_SAVE_AFTER_BATTLE_STABLE_TICKS) {
```
LordMonoxide: Instead of waiting for a certain number of ticks, I think it would be better to just do the save here:

Image

Can also remove the autoSaveAfterBattlePending var and just check it in that method. Maybe make a new method called `autoSaveIfEnabled` or whatever to keep things tidy.

src/main/java/legend/game/submap/SMap.java
```diff
@@ -541,10 +559,52 @@ public boolean canSave() {
...
+    return this.smapLoadingStage_800cb430 == SubmapState.RENDER_SUBMAP_12
+      && whichMenu_800bdc38 == WhichMenu.NONE_0
+      && !gameState_800babc8.indicatorsDisabled_4e3
+      && !this.transitioning_800f7e4c
+      && fullScreenEffect_800bb140.currentColour_28 == 0
+      && this.canSave();
+  }
+
+  private void createNewSave() {
+    try {
```
LordMonoxide, line 566: `GameOverlay.addNotification`?

LordMonoxide (changes requested):

src/main/java/legend/game/submap/SMap.java
Ink230: removed audio notice, added the notification

src/main/java/legend/game/submap/SMap.java
Ink230: agreed

src/main/java/legend/game/submap/SMap.java
Ink230: done

LordMonoxide (approved):

PR #2789, opened 2026-08-27
fix AudioThread track OS default for unset audio device configuration

Ink230:
### Motivation

Bluetooth headphones and one Speaker, no other audio devices. Turning bluetooth off or turning the bluetooth device off force changes windows default device to the Speaker.

SC I noticed was not auto changing its audio output when the Audio config is set to `Default`.

### Description

- Default is a special option that will now track and always use whatever the Default device is set to on the OS
- Specific device selections will ignore this new behaviour as usual

### DIFF
--- a/src/main/java/legend/core/audio/AudioThread.java
+++ b/src/main/java/legend/core/audio/AudioThread.java
@@ -270,7 +270,9 @@ public final class AudioThread implements Runnable {
         }
 
         if(this.deviceChanged) {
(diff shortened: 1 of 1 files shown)


LordMonoxide:
I'm confused, this looks like it doesn't change the behaviour of the `<default>` option, which in my experience was working. How are you getting a configured device that's an empty string?

LordMonoxide:
I looked into this a bit more, it's an issue in AudioDeviceConfig. The default value is in fact an empty string. This is fine.

PR #2790, opened 2026-08-27
Add UI text box colour settings preview and randomization buttons

Ink230:
https://streamable.com/huy2sh


Ink230:
image
Fixed arrow placements from video

PR #2797, opened 2026-08-30
New tagged save format

LordMonoxide:
This will make it must easier to add to or change the save format in the future


src/main/java/legend/game/types/GameState52c.java
```diff
@@ -74,8 +74,8 @@ public class GameState52c {
 //  public boolean vibrationEnabled_4e1;
   public boolean indicatorsDisabled_4e3;
 
-  /** A bitset used to set each char's MP to max the first time each one is loaded */
-  public int characterInitialized_4e6;
+//  /** A bitset used to set each char's MP to max the first time each one is loaded */
+//  public int characterInitialized_4e6;
```
LordMonoxide, line 78: Unused, this was moved to `CharacterData2c` when I added templates

src/main/java/legend/game/submap/SMap.java
```diff
@@ -454,63 +456,53 @@ public void destroy() {
...
   }
 
   @Override
-  public void readSaveData(final GameState52c gameState, final FileData data) {
+  public void readSaveData(final GameState52c gameState, @Nullable final Tag tag) {
     this.unstuckMovementBudget = UNSTUCK_SAVE_LOADED_BUDGET;
 
     // no data - legacy saves
-    if(data.size() == 0) {
+    if(tag == null) {
```
avionanx, line 483: Pulled code and loaded an older save on wmap, transitioned to submap and game crashed, submapscene and cut were not set. This is null only when I loaded from wmap. I think EngineStates.engineStateData is cleared when game loads on wmap?

PR #2801, opened 2026-09-02
Implemented support for alternative languages

dougnottdougie:
French discs work now. Drop the SCES ISOs in isos/ and it detects them, unpacks to files/fr and uses the French charset. US unpacks to files/us, and an old bare files/ gets moved there on first run.

Everything that was hardcoded to the US build (disc IDs, exe name, engine overlay offsets, charset) lives in GameRegion now, so another region is one enum entry.

Gaps: PAL FMVs are skipped, the MDEC demuxer overruns on them. 20 of the 347 script patches don't apply to the French scripts, they're diffs against the English disassembly. The font atlases had no accents at all, so there are 23 new glyphs below the existing ones.

Three of the fixes in here are US bugs too: the engine overlay slice length was hardcoded even though the decompressor stops on its own, a missing intro.mp4 gave a black screen forever, and a 0x0 FMV frame killed the render thread. Happy to split those out if you'd rather.


dougnottdougie:
image
image
French successfully pulling in dialogue. Its not perfect, but its a start.

LordMonoxide:
There are many issues with this, I don't have time to fix LLM code

Issue #2809, opened 2026-09-06
Enhancement: utilize hasGood instead of flags for various goods


pkolb-dev:
Creating this issue to use the GH# issue number for the scripts.csv.

piggy-backing off of #2794, this issue will track the other goods that do something similar instead of checking for the physical good.

The following locations will be addressed here:

The prairie tree does not recognize if the player has the good "Axe from the Shack". This is an enhancement to utilize SItem::scriptHasGood to do the check rather than rely on story flags. 

water in dragon nest does not recognize if the player has the good "Water Bottle". 

the strange plant that blocks off shrine of shirley does not recognize if the player has the good "Life Water"

The boat in Furni does not recognize if the player has the good "Boat license"

The intent of the fixes will be to ensure that the new behavior matches the behavior of black castle's Stones and Magic Oil.

pkolb-dev:
#2821 is preferred instead of editing each script file.

LordMonoxide:
Closed by #2821

PR #2817, opened 2026-09-10
Feature: Utilize hasGood for axe from shack

pkolb-dev:
Part of #2809 

I elected to check the good on the last bit of the tree's interaction. This way, players still need to trigger the proper story flags to enable the hasGood check.


pkolb-dev:
Does not close the issue for now

PR #2821, opened 2026-09-10
Add new event for script flags

pkolb-dev:
In the need for archipelagoon to create consistent, logical blockers based off of goods held I created this event.

Whenever the submap scripts try to read global flags they will fire an event that allows mods to "alter" the read value.

In my case, I can do something like this for utilizing the LifeWater good:
```
@EventListener
  public void readScriptFlags(final ReadGlobalFlagsEvent event) {
    if(event.getFlagIndex() == 59 && event.flagArray == ScriptFlagArrayEnum.FLAGS2) {
      event.flagValue = gameState_800babc8.goods_19c.has(GameEngine.REGISTRIES.goods.getEntry(LodGoods.LIFE_WATER.getId()));
    }
  }
```

The code snippet looks for the specific flag index in the flags2 array and rather than relying on story flags, it'll utilize the boolean logic in `goods_19c.has`.

Note: this event fires A LOT.


pkolb-dev:
Closes: #2809

LordMonoxide (approved):

